PR #2, opened 2016-05-11
Add setting for publishing on Sonatype

rafaparadela:
This PR, not only adds settings to configure publishing snapshots but also it helped me to understand how it works. Fancy documentation is almost ready.

@FPerezP @fedefernandez @javipacheco Could you review please? Thank you in advance.


build.sbt
```diff
@@ -2,7 +2,7 @@ organization := "com.fortysevendeg"
 
 name := "github4s"
 
-version := "0.1"
+version := "0.1-SNAPSHOT"
```
javipacheco: We should create  `0.1` version and later we can create a new PR with `0.1-SNAPSHOT`

publish.sbt
```diff
@@ -0,0 +1,54 @@
+organizationName := "47 Degrees"
+
+organizationHomepage := Some(new URL("http://47deg.com"))
+
+publishMavenStyle := true
+
+startYear := Some(2015)
```
javipacheco: 2016?

build.sbt
rafaparadela: Makes sense

publish.sbt
rafaparadela: Done ;-)

javipacheco:
LGTM!

PR #3, opened 2016-05-12
Remove unused implementation

rafaparadela:
Since we don't need to split GHResult, this PR simplifies how the response is returned. Some unused algebras have been removed as well.

@FPerezP Could you review please?


FPerezP:
Looks good.

PR #5, opened 2016-05-12
Fixes test

rafaparadela:
This PR fixes tests. It's get an accessToken from an environment variable in order to avoid the limit of unauthenticated calls.

@FPerezP @javipacheco Could you take a look please? Thank you.


javipacheco:
God job! LGTM!

PR #9, opened 2016-05-16
ScalaDoc

rafaparadela:
This PR includes comments in order to generate Scala API Doc. 

@javipacheco Could you take a look please?

### DIFF
--- a/src/main/scala/com/fortysevendeg/github4s/api/Auth.scala
+++ b/src/main/scala/com/fortysevendeg/github4s/api/Auth.scala
@@ -10,6 +10,7 @@ import io.circe.generic.auto._
 import io.circe.syntax._
 import scalaj.http.HttpConstants._
 
+/** Factory to encapsulate calls related to Auth operations  */
 object Auth {
 
   protected val httpClient = new HttpClient()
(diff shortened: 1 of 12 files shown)


src/main/scala/com/fortysevendeg/github4s/api/Auth.scala
```diff
@@ -21,13 +22,12 @@ object Auth {
...
-    * @param note
-    * @param client_id
-    * @param client_secret
-    * @return
+    * @param username the username of the user
+    * @param password the password of the user
+    * @param scopes attached to the token
+    * @param note to remind you what the OAuth token is for
+    * @param client_id the 20 character OAuth app client key for which to create the token
+    * @param client_secret the 40 character OAuth app client secret for which to create the token
```
javipacheco: Maybe we should include return type here

```
@return GHResponse[Authorization] comment
```

src/main/scala/com/fortysevendeg/github4s/api/Auth.scala
```diff
@@ -46,10 +46,9 @@ object Auth {
   /**
     * Generates the authorize url with a random state, both are returned within Authorize object
     *
-    * @param client_id
-    * @param redirect_uri
-    * @param scopes
-    * @return
+    * @param client_id the 20 character OAuth app client key for which to create the token
+    * @param redirect_uri the URL in your app where users will be sent after authorization
+    * @param scopes attached to the token
```
javipacheco: See above

javipacheco:
I think that we should use `@return` in all calls and the coverage doesn't work fine 

`coverage is n/a%`

src/main/scala/com/fortysevendeg/github4s/api/Auth.scala
rafaparadela: Ok

rafaparadela:
- `@return` annotation: OK, makes sense
- The coverage report locally is: 71.31% (I don't know why Codecov.io doesn't infer the right value.

http://scala.run/1TjEeTf

rafaparadela:
@javipacheco Your comments have been addressed. I'm going to increase test coverage in next PRs, is good to merge so far?

javipacheco:
LGTM

PR #13, opened 2016-05-23
Moved/Renamed package in order to omit organization prefix

rafaparadela:
This PR:
- Moves the package `github4s` out of the path structure `com/fortysevendeg`.
- Renames packages in order to omit the prefix `com.fortysevengdeg`.

@dialelo Could you review please? Thanks for the suggestion.


dialelo:
LGTM :shipit:

PR #18, opened 2016-05-30
Provide interpreters with several monad instances

rafaparadela:
This PR:
- Provides implicit interpreters for `cats.Id`, `cats.Eval`, `Future` and `scalaz.Task`.
- Isolates `scalaz` dependency in other module.
- Adds examples of use in documentation (tut).
- Removes `com.typesafe.config` dependency.

@dialelo @raulraja Could you review please? Thank you.


docs/src/jekyll/_config.yml
```diff
@@ -3,6 +3,7 @@ description: "A GitHub API wrapper written in Scala"
 github_owner: 47deg
 baseurl: /github4s
 style: github4s
+highlight_theme: tomorrow
```
raulraja: Nice!

docs/src/tut/docs.md
```diff
@@ -2,6 +2,82 @@
...
+val u3 = Github().users.get("dialelo").exec[Future]
+Await.result(u3, 2.seconds)
+```
+
+WIP: With `scalaz.Task`
+
+```tut:silent
+import scalaz.concurrent.Task
+import github4s.scalaz.Implicits._
+import github4s.scalaz.Implicits.TaskInterpreters._
```
raulraja: Lowercase implicits.see cats lib for names around implícits, syntax, etc... Also shouldn't the implícits import already bring into scope the Task interpreter?

src/main/scala/github4s/Implicits.scala
```diff
@@ -0,0 +1,63 @@
...
+      Eval.later({
+        try {
+          fa.value
+        } catch {
+          case e: Throwable ⇒ f(e).value
+        }
+      })
+  }
+
+  implicit def idMonadError(implicit I: Monad[Id]): MonadError[Id, Throwable] = new MonadError[Id, Throwable] {
```
dialelo: Note that the instance of `MonadError[Id, Throwable]` doesn't fullfill the laws of MonadError since exceptions cannot be captured. Do we want to provide it as part of the library knowing this?

raulraja: Have you tested I'd and task in a file where either one was required as implícits. @dialelo and I have experienced implícits ambiguity presumably because Task also conforms to Id given I'd is applicable to anything. Not sure we should provide this instance here. We may have to look into a LowerPriorityImplicits trait stack to avoid ambiguity. See also cats for examples of this.

src/test/scala/github4s/utils/DummyGithubUrls.scala
```diff
@@ -0,0 +1,12 @@
+package github4s.utils
+
+import github4s.GithubApiUrls
+
+trait DummyGithubUrls {
+
+  implicit val dummyUrls = GithubApiUrls(
```
raulraja: All implícits should have a explicit return type.

scalaz/src/main/scala/github4s/scalaz/Implicits.scala
```diff
@@ -0,0 +1,25 @@
+package github4s.scalaz
+
+import cats.MonadError
+import github4s.free.interpreters.Interpreters
+import scalaz.concurrent.Task
+
+object Implicits {
+
+  implicit val evalMonadError: MonadError[Task, Throwable] = new MonadError[Task, Throwable] {
```
dialelo: The name should be `taskMonadError`. Note that implicits can clash by name, so maybe we want to uniquify the names of the implicits throughout the library. In this case we could use `github4sTaskMonadError` for example.

scalaz/src/main/scala/github4s/scalaz/Implicits.scala
```diff
@@ -0,0 +1,25 @@
...
+
+    override def pure[A](x: A): Task[A] = Task.now(x)
```
dialelo: I'd overwrite `pureEval` too for using `Task.now` when the value has been already evaluated, like [they do in Cats with Future](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/std/future.scala#L16-L19). This avoids introducing an unneeded asynchronous boundary.

dialelo:
Awesome work!

rafaparadela:
Comments have been addressed. 
@raulraja @dialelo does the code look good to you? Than you.

src/main/scala/github4s/Implicits.scala
```diff
@@ -0,0 +1,62 @@
+package github4s
+
+import cats.std.FutureInstances
+import cats.std.future._
+import cats.{ Monad, Id, Eval, MonadError }
+import github4s.free.interpreters.Interpreters
+import scala.concurrent.{ ExecutionContext, Future }
+
+object implicits extends Interpreters with EvalInstances with IdInstances with FutureInstances {
```
dialelo: We can drop the braces here.

dialelo:
:shipit: LGTM

src/main/scala/github4s/Implicits.scala
rafaparadela: Makes sense

PR #28, opened 2016-06-29
Take into account that commit authors may be null

dialelo:
Some users don't properly configure their git email to be the same as
their GitHub email, thus GitHub can't figure out which author (on
GitHub) a commit has. I've modified Commit and its decoder to take this
into account adding `Option` around the fields that can be null.


dialelo:
@raulraja @rafaparadela plz take a look.

raulraja:
LGTM!

PR #46, opened 2016-12-19
Adds crossover for Scala 2.10 and upgrades libraries

fedefernandez:
This PR:

* Adds crossover for Scala 2.10. Because the `case class` `ReposUrls` had more than 22 fields, I've replaced all optional fields for a map, adding compatibility with scala 2.10
* Upgrades the `sbt-catalysts-extras` plugin. This implies some changes in the cats library code, like replacing `Xor` by `Either` and avoid some deleted implicits (`RecursiveTailRecM`)

Please @juanpedromoreno or @jdesiloniz could you review? Thanks

**NOTE** Should I change the version to `0.10.0-SNAPSHOT`? I think we are breaking the compatibility with the 0.9.x version changing the `ReposUrls` case class.


build.sbt
```diff
@@ -19,6 +19,11 @@ lazy val buildSettings = Seq(
     homepage := Option(url("http://47deg.github.io/github4s/")),
     organizationHomepage := Option(new URL("http://47deg.com")),
     scalaVersion := "2.11.8",
+    crossScalaVersions := Seq("2.10.6", scalaVersion.value),
```
juanpedromoreno: Could you please solve this issue by the way?
https://github.com/47deg/github4s/issues/40

`crossScalaVersions := Seq("2.10.6", "2.11.8", "2.12.1")` would be great ;) . Thanks!

fedefernandez: There are some incompatibilities with the `RosHTTP` library ([see this build](https://travis-ci.org/47deg/github4s/jobs/185127926)). I'm going to send a PR for adding support to Scala 2.12 in `RosHTTP` but in the meantime, I think we should go forward with this. What do you think?

juanpedromoreno: 👍

PR #95, opened 2017-04-24
Avoids executing create and edit issue operations

fedefernandez:
Please @juanpedromoreno could you review? Thanks


BenFradet:
I'll make sure to avoid `tut:silent` for the other examples that might be problematic :+1:

PR #101, opened 2017-04-25
Replace foldLeft with traverse in Decoders

peterneyens:


github4s/jvm/src/test/scala/github4s/unit/DecodersSpec.scala
```diff
@@ -50,4 +51,12 @@ class DecodersSpec extends FlatSpec with Matchers with FakeResponses {
...
   }
 
+  "NonEmptyList Decoder" should "return a valid NonEmptyList for a valid JSON" in {
+    decode[NonEmptyList[Int]]("[1,2,3]") should be(Right(NonEmptyList.of(1, 2, 3)))
+  }
+
+  it should "return an error for an empty list" in {
+    decode[NonEmptyList[Int]](emptyListResponse) should be('left)
+  }
+
```
BenFradet: I think it might be worth it to add a test case for a single element because I think that's why this decoder is here for in the first place.

Something like:

```scala
case class Foo(a: Int)
it should "return a valid NonEmtpyList for a valid JSON" in {
  decode[NonEmptyList[Foo]]("""{"a": 1}""") should be(Right(NonEmptyList.of(Foo(1))))
}
```

What do you think?

peterneyens: Hmm this seems to be a bit dodgy decoder, converting both `A` and `[A]`, but as that seems what is needed, I'll add that test :+1:

BenFradet: agreed, the github api and its quirks ...

PR #126, opened 2017-05-09
Streamlined the getting started

BenFradet:


docs/src/main/tut/docs.md
```diff
@@ -5,135 +5,142 @@ title: Getting Started
...
 
-To get an user
+As an introductory example, we can get a user with the following:
 
 ```tut:silent
 val user1 = Github(accessToken).users.get("rafaparadela")
 ```
 
-user1 in this case `Free[GHException Xor GHResult[User], User]` and we can run (`foldMap`) with `exec[M[_], C]` where `M[_]` represent any type container that implements `MonadError[M, Throwable]`, for instance `cats.Eval`; and C represents a valid implementation of an HttpClient. The previously mentioned implicit classes carry already set up instances for working with `scalaj` (for JVM-compatible apps) and `roshttp` (for scala-js-compatible apps). Take into account that in the latter case, you can only use `Future` in the place of `M[_]`:
+`user1` in this case `Free[GHException Xor GHResult[User], User]` which we can run (`foldMap`) with
```
juanpedromoreno:
```diff
- `user1` in this case `Free[GHException Xor GHResult[User], User]` which we
+ `user1` in this case `Free[GitHub4s, Either[GHException, GHResult[User]]]` which we
```

juanpedromoreno: Actually, it's a `GHIO[GHResponse[User]]`, if we consider type aliases:

```
type GHIO[A] = Free[GitHub4s, A]
type GHResponse[A] = Either[GHException, GHResult[A]]
```

BenFradet:
@juanpedromoreno Thanks for your review, I think I've clarified things in 84b7246 :+1:

BenFradet:
@fedefernandez good to go :+1:

PR #138, opened 2017-05-18
Contributing guide

BenFradet:


docs/src/main/tut/contributing.md
```diff
@@ -0,0 +1,386 @@
...
+      // ...
+      case ListStatuses(owner, repo, ref, accessToken) =>
+        repos.listStatus(accessToken, headers, owner, repo, ref)
+      // ...
+    }
+  }
+}
+```
+
+## Wrapper
```
BenFradet: I don't what happened but the `Wrapper` section is offset almost centered.

docs/src/main/tut/contributing.md
```diff
@@ -0,0 +1,386 @@
+---
+layout: docs
```
juanpedromoreno: I'm wondering if it makes sense to put this awesome contributing guide in the microsite main page, using `page` layout. We did it in cats in this way: http://typelevel.org/cats/contributing.html

What do you think?

docs/src/main/tut/contributing.md
```diff
@@ -0,0 +1,386 @@
...
+
+## Token
+
+The first step needed to run the tests is a valid token which we can provide through an
+environment variable:
+
+```bash
+export GITHUB4S_ACCESS_TOKEN=aaaa
+```
+
```
juanpedromoreno: I'd add a link indicating how generate this personal access token

docs/src/main/tut/contributing.md
```diff
@@ -0,0 +1,386 @@
...
+## Integration tests
+
+The integration tests are grouped by API in [github4s.integration package][integ-pkg]. As a result,
+we'll be writing our tests in [GHReposSpec][repos-integ-spec-scala]:
```
juanpedromoreno: Let's add that we only want integration tests for read-only operations

BenFradet: the funny thing is that I was planning on adding it but completely forgot about it :smile:

docs/src/main/tut/contributing.md
BenFradet: Good idea, that way we avoid cluttering the documentation too much.

BenFradet: Should we move it to a `CONTRIBUTING.md` at the root of the repo as in cats?

juanpedromoreno: We could, however, it's not mandatory. If we want to move it to the root I'm fine with that. Nevertheless, I'd separate it from the current `CONTRIBUTING.md` file in the root. What do you think about `DEV_GUIDE.md` or something similar? (afterward, when it's copied to the microsite territory we can rename it on the fly).

BenFradet: I'm fine with leaving it where it is for now.

PR #143, opened 2017-05-22
Microsite Enhancements

juanpedromoreno:
This PR brings some microsite improvements:

* Proper attributions in the global footer, where, from now on, it'll show `Github4s is designed and developed by Github4s contributors` pointing to the AUTHORS file.
* Extra explanation saying g4s is compatible for both Scala JVM/JS and Scala 2.10, 2.11 and 2.12.
* Adds changelog as link, which might be handy. It has been placed side to the `Contributing` section.


juanpedromoreno:
Wait until https://github.com/47deg/github4s/pull/144 is merged.

PR #144, opened 2017-05-22
Fixes ghost users associated with pull requests, issues and comments

juanpedromoreno:
When one user is removed completely from Github, the API returns null (https://github.com/ghost), a possibility not contemplated until now.

This PR makes the users associated with issues, pull requests and issue comments as an optional field.


juanpedromoreno:
👍 

* In the first case, I don't think either.
* In the release case, I've just tested it and it turns out that the API doesn't return any result (404) for tags without authors.
* For the statuses, here the most interesting result [500 error 😮  ](https://github.com/47deg/github4s/commits/master?after=aaee622ee5c8c8b4256fc20be7322bf67f05701d+104)

juanpedromoreno:
Anyway, let's make optional releases and statuses.

PR #153, opened 2017-05-26
Add the merges endpoint under GHRepos.

mscharley:
Another WIP PR. Works for the known-good use case. [Documentation](https://developer.github.com/v3/repos/merging/#perform-a-merge)

Known issues:

* [ ] A 204 response which means that the merge would be a NOP will generate a `Left(github4s.GithubResponses$JsonParsingException: exhausted input)` instead of a useful error.

### DIFF
--- a/github4s/shared/src/main/scala/github4s/GithubAPIs.scala
+++ b/github4s/shared/src/main/scala/github4s/GithubAPIs.scala
@@ -109,6 +109,15 @@ class GHRepos(accessToken: Option[String] = None)(implicit O: RepositoryOps[GitH
       context: Option[String] = None
   ): GHIO[GHResponse[Status]] =
     O.createStatus(owner, repo, sha, state, target_url, description, context, accessToken)
+
+  def merge(
+      owner: String,
+      repo: String,
+      base: String,
+      head: String,
+      commitMessage: Option[String] = None
+  ): GHIO[GHResponse[MergeResponse]] =
+    O.merge(owner, repo, base, head, commitMessage, accessToken)
 }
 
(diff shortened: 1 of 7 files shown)


mscharley:
Possible solution to the 204 issue may be to return an `Option[MergeResponse]`, but I'm not sure that's a good encoding for the NOP case.

juanpedromoreno:
@mscharley An `Option[MergeResponse]` sounds reasonable to me.

mscharley:
Next problem is that Circe refuses to decode an empty string to anything because it immediately runs out of input to deserialise leading to the `exhausted input` error above. Not sure where/how to approach this problem at the moment.

juanpedromoreno:
`Either[Unit, MergeResponse]`?

Potentially we'd need a special run extension (`runEither` perhaps?). Dummy example:

```scala
import io.circe._, io.circe.generic.auto._, io.circe.parser._, io.circe.syntax._
import cats.syntax.either._

val emptyResponse = ""

val nonEmptyResponse = """
      |{
      |  "foo": "bar"
      |}
      """.stripMargin

parse(emptyResponse).leftMap(_ => ())
// res4: Either[Unit, Json] = Left(())
parse(nonEmptyResponse).leftMap(_ => ())
// res5: Either[Unit, Json] = Right(JObject(MapAndVectorJsonObject(Map("foo" -> JString("bar")), Vector("foo"))))
```

That's what I can think now, likely there will be a better solution.

mscharley:
Yeah, the completely non existent response entity kinda messes with the
fact that Circe is a requirement for the HTTP response entity at the moment.

PR #155, opened 2017-05-30
cats-effect module

BenFradet:
- [x] js support
- [x] remove hardcoded cats-effect version once sbt-org-policies is out
- [x] doc


cats-effect/src/main/scala/github4s/cats/effect/Implicits.scala
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package github4s.cats.effect
+
+import cats.effect.IO
+import github4s.HttpRequestBuilderExtensionJVM
+import github4s.free.interpreters.{Capture, Interpreters}
+import scalaj.http.HttpResponse
+
+object implicits extends HttpRequestBuilderExtensionJVM {
```
BenFradet: should I create one implicits for the jvm and one for js?

build.sbt
```diff
@@ -39,3 +39,8 @@ lazy val scalaz = (project in file("scalaz"))
   .settings(moduleName := "github4s-scalaz")
   .settings(scalazDependencies: _*)
   .dependsOn(github4sJVM)
+
+lazy val catsEffect = (project in file("cats-effect"))
+  .settings(moduleName := "github4s-cats-effect")
+  .settings(catsEffectDependencies: _*)
+  .dependsOn(github4sJVM)
```
BenFradet: is it ok to depend on both `github4sJVM` and `github4sJS`?

project/ProjectPlugin.scala
```diff
@@ -71,8 +71,14 @@ object ProjectPlugin extends AutoPlugin {
...
     lazy val docsDependencies: Def.Setting[Seq[ModuleID]] = libraryDependencies += %%("scalatest")
 
-    lazy val scalazDependencies: Def.Setting[Seq[ModuleID]] = libraryDependencies += %%(
-      "scalaz-concurrent")
+    lazy val scalazDependencies: Def.Setting[Seq[ModuleID]] =
+      libraryDependencies += %%("scalaz-concurrent")
+
+    lazy val catsEffectDependencies: Def.Setting[Seq[ModuleID]] =
+      libraryDependencies ++= Seq(
+        "org.typelevel" %% "cats-effect" % "0.3",
```
BenFradet: to rm

build.sbt
juanpedromoreno: I don't know if `cat-effects` is already a crossed-project with Scalajs (but it will be likely). If that's already the case, we could make `catsEffect` module as `crossProject` too.

cats-effect/src/main/scala/github4s/cats/effect/Implicits.scala
juanpedromoreno: By now, I think that you should. Nevertheless, I guess we want to get rid of this kind of implicit import.

https://github.com/47deg/github4s/issues/45

One option would use a fully compatible http client for both jvm and js: https://github.com/pepegar/hammock . 

What do you think?

project/ProjectPlugin.scala
juanpedromoreno: `sbt-org-policies 0.5.2` has been already released: https://github.com/47deg/sbt-org-policies/releases/tag/v0.5.2

build.sbt
BenFradet: yes it supports scala js, I'll try making it a cross project

cats-effect/src/main/scala/github4s/cats/effect/Implicits.scala
BenFradet: in the long run I think this would be great indeed.

juanpedromoreno:
BTW I don't know why my review was sent twice XD

cats-effect/src/main/scala/github4s/cats/effect/Implicits.scala
BenFradet: welp I think I see what you mean since roshttp is tied to Future we can't support IO in scala js

BenFradet:
Since it seems we can't use a `Capture` with `HttpRequestBuilderExtensionJS`, we can either:

- build a `HttpRequestBuilderExtensionJS` specifically for `IO` in cats-effect js
- not support cats-effect in js for the moment (until we port everything to hammock)

What do you think?

juanpedromoreno:
Makes sense  :thumbsup:

cats-effect/js/src/test/scala/github4s/cats/effect/js/CatsEffectJSSpec.scala
```diff
@@ -0,0 +1,51 @@
...
+
+  it should "return a failed result for an invalid call" in {
+    val response = Github(accessToken).users
+      .get(invalidUsername)
+      .exec[IO, SimpleHttpResponse](headerUserAgent)
+
+    val res = response.unsafeRunSync
+    res.isLeft shouldBe true
+  }
+}
```
BenFradet: This spec is never run, I don't know why.

juanpedromoreno: See https://github.com/47deg/github4s/pull/156

docs/src/main/tut/docs.md
```diff
@@ -125,6 +124,33 @@ object ProgramTask {
 
 Note that you'll need a dependency to the `github4s-scalaz` pacakge to leverage `scalaz.Task`.
 
+### Using `cats.effect.IO`
+
+On the JVM:
+```scala
```
BenFradet: I can't seem to get this example to compile with tut despite adding cats-effect to the doc deps, any ideas?

juanpedromoreno: See https://github.com/47deg/github4s/pull/158

project/ProjectPlugin.scala
```diff
@@ -103,9 +109,9 @@ object ProjectPlugin extends AutoPlugin {
       orgSupportedScalaJSVersion := Some("0.6.15"),
       orgScriptTaskListSetting ++= List(
         (ScoverageKeys.coverageAggregate in Test).asRunnableItemFull,
-       "docs/tut".asRunnableItem
+        "docs/tut".asRunnableItem
       ),
-      coverageExcludedPackages := "<empty>;github4s\\.scalaz\\..*",
+      coverageExcludedPackages := "<empty>;github4s\\.scalaz\\..*;github4s\\.cats\\.effect\\..*",
```
juanpedromoreno: Why are we excluding these tests?

BenFradet: Just thought I'd exclude them like what is being done for scalaz, happy to include them.

juanpedromoreno: 👍 

scalaz doesn't have tests, cats-effect has thanks to your work :)

BenFradet:
mmh I get 100% coverage on cats-effect-jvm but 78.57% on js, because of the HttpRequestBuilder for IO, I'll try adding tests for that as well.

juanpedromoreno:
Ohh, thanks. Let's try it, if not, we could remove the 80% restriction.

BenFradet:
Yup, I'll try tomorrow.

Anyway the low coverage is only temporary since once we move to hammock there'll be no HttpRequestBuilder extension for IO and the capture instance will be used in JS so we'll be at 💯 for both JVM and JS.

BenFradet:
@juanpedromoreno I think it's ready for a review :+1:

BenFradet:
Thanks a lot for your help, merging!

PR #156, opened 2017-06-06
Fixes cats-effects ScalaJS tests

juanpedromoreno:
This PR solves the sbt configuration to allow run tests for scalajs cross building (`crossDepSettings `).

On the other hand, there was a problem in tests because:
`java.lang.UnsupportedOperationException: cannot synchronously await result on JavaScript; use runAsync or unsafeRunAsync`

Hence, I also fixed both existing tests with the recommended way to do it.

Feel free to merge it if this makes sense to you.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -41,7 +41,7 @@ lazy val scalaz = (project in file("scalaz"))
 
 lazy val catsEffect = (crossProject in file("cats-effect"))
   .settings(moduleName := "github4s-cats-effect")
-  .settings(catsEffectDependencies: _*)
+  .crossDepSettings(catsEffectDependencies: _*)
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -41,7 +41,7 @@ lazy val scalaz = (project in file("scalaz"))
 
 lazy val catsEffect = (crossProject in file("cats-effect"))
   .settings(moduleName := "github4s-cats-effect")
-  .settings(catsEffectDependencies: _*)
+  .crossDepSettings(catsEffectDependencies: _*)
```
BenFradet: aaah :+1:

juanpedromoreno:
👍

PR #162, opened 2017-09-12
Support for the list organization repositories endpoint

BenFradet:


BenFradet:
@juanpedromoreno woops thanks for calling it out, I had completely forgotten about the doc in this one.

Will add it to his PR :+1:

PR #164, opened 2017-09-17
Bump sbt-org-policies to 0.7.4

BenFradet:
It also migrates from cats `0.9.0` to `1.0.0-MF`, among other upgrades and enhancements.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,2 +1,3 @@
+addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.7.0")
 addSbtPlugin(
-  "com.47deg" % "sbt-org-policies" % "0.5.4" exclude ("io.get-coursier", "sbt-coursier"))
+  "com.47deg" % "sbt-org-policies" % "0.7.4" exclude ("io.get-coursier", "sbt-coursier"))
--- a/github4s/shared/src/main/scala/github4s/api/package.scala
+++ b/github4s/shared/src/main/scala/github4s/api/package.scala
@@ -21,6 +21,6 @@ import io.circe.{Json, Printer}
 package object api {
 
   def dropNullPrint(json: Json): String =
-    Printer.noSpaces.copy(dropNullKeys = true).pretty(json)
+    Printer.noSpaces.copy(dropNullValues = true).pretty(json)
 
 }
--- a/github4s/shared/src/main/scala/github4s/app.scala
+++ b/github4s/shared/src/main/scala/github4s/app.scala
@@ -16,16 +16,16 @@
 
 package github4s
 
-import cats.data.Coproduct
+import cats.data.EitherK
 import github4s.free.algebra._
 
 object app {
-  type COGH01[A]   = Coproduct[RepositoryOp, UserOp, A]
-  type COGH02[A]   = Coproduct[GistOp, COGH01, A]
-  type COGH03[A]   = Coproduct[IssueOp, COGH02, A]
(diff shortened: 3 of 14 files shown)


BenFradet:
Thanks, I wasn't entirely sure sbt-org-policies needed to be involved, I'll take care of https://github.com/47deg/sbt-org-policies/issues/538 as well then.

juanpedromoreno:
@BenFradet I needed to upgrade sbt-org-policies and btw I upgraded everything in general: https://github.com/47deg/sbt-org-policies/pull/632, including Scalajs.

BenFradet:
Thanks a lot! I'll bump sbt-org later today :+1:

BenFradet:
mmmh getting https://travis-ci.org/47deg/github4s/jobs/276970366#L846 with 0.6.4 and sbt doesn't launch with 0.7.0

suhasgaddam:
@BenFradet I think Scala.js `0.6.20` requires sbt `0.13.16`.
with
`sbt-org-policies`:`0.7.0`
`build.properties`: `sbt.version = 0.13.16`

`sbt +compile` seems to work for me, locally

Links that I stumbled on and for anyone else searching:
https://stackoverflow.com/questions/46206462/java-lang-nosuchmethoderror-sbt-package-zerolsbt-global
http://www.scala-js.org/news/2017/09/01/announcing-scalajs-0.6.20/ -> `Breaking Changes`

suhasgaddam:
The travis ci error: `sbt-org-policies`:`0.6.4` did not have `sbt-microsites` enabled. We are/were still working on getting `sbt-microsites` cross-compiling with sbt 1.0.

BenFradet:
@suhasgaddam thanks! didn't know about the issue with microsites / 0.13.16 min for 0.6.20

project/plugins.sbt
```diff
@@ -1,2 +1,3 @@
+addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.7.0")
 addSbtPlugin(
-  "com.47deg" % "sbt-org-policies" % "0.5.4" exclude ("io.get-coursier", "sbt-coursier"))
+  "com.47deg" % "sbt-org-policies" % "0.6.4" exclude ("io.get-coursier", "sbt-coursier"))
```
juanpedromoreno: It should be `0.7.0`

BenFradet: fixed :fearful:

BenFradet:
Weirdly enough it seems to compile fine even though `cats.data.Coproduct` was removed from cats-1.0.0-MF.

However, it fails during the cats-effect tests with https://travis-ci.org/47deg/github4s/jobs/277197841#L2596.

Still investigating :confused:

BenFradet:
@juanpedromoreno fixed! :+1: Regarding releasing a version, do I just need to edit `version.sbt`? Should I introduce a specific commit?

juanpedromoreno:
Yes, you only need to modify the `version.sbt`.

Are we releasing the `0.1.5.1` or `0.16.0`? As it's not binary compatible I propose `0.16.0`, what do you think?

Either way, I'd suggest doing it in a different PR, since due to https://github.com/47deg/sbt-org-policies/issues/637, the plugin is not running the release task correctly. Hence, we might need to edit the `orgAfterSuccessCI` task by something like this https://github.com/47deg/sbt-org-policies/blob/2c89b98d9366b626a978e3ff17cbc25e561ea48b/.travis.yml#L31-L44, in the `.travis.yml` file.

BenFradet:
@juanpedromoreno ok thanks :+1: , I'll open a new PR

PR #170, opened 2017-10-12
update build

AdrianRaFo:
Update to sbt 1.0.x

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: scala
 dist: precise
 
 scala:
-- 2.10.6
 - 2.11.11
 - 2.12.3
 
--- a/build.sbt
+++ b/build.sbt
@@ -3,7 +3,6 @@ pgpPublicRing := file(s"$gpgFolder/pubring.gpg")
 pgpSecretRing := file(s"$gpgFolder/secring.gpg")
 
 lazy val root = (project in file("."))
-  .dependsOn(github4sJVM, github4sJS, scalaz, catsEffectJVM, catsEffectJS, docs)
   .aggregate(github4sJVM, github4sJS, scalaz, catsEffectJVM, catsEffectJS, docs)
   .settings(noPublishSettings: _*)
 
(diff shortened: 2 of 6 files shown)


juanpedromoreno:
Looks good, but builds seem to be failing 🤔

AdrianRaFo:
I try to run the tests in master and failed too, integrations test by the way, so it can be failing by the test repo, @juanpedromoreno can you confirm it?

juanpedromoreno:
Mmm... It seems to be a binary incompatibility, I'd need to look at it deeper

juanpedromoreno:
Is the current issue the same that you found last week, @suhasgaddam ?

AdrianRaFo:
I'm going to mark this as blocked until the test library is fixed

juanpedromoreno:
Could we go ahead keeping `0.13.16`?

AdrianRaFo:
I can try but freestyle doesn't is for scala 2.10, It's ok if I remove that support?

suhasgaddam:
@AdrianRaFo unless I missed a PR, github4s still supports 2.10 and has not switched to freestyle yet. The build project can use 0.13/2.10, and the "real" project can use any Scala version. (I think I am using bad wording with respect to build vs "real" project.)

AdrianRaFo:
@suhasgaddam the purpose of my pr is to prepare github4s to make a freestyle migration

BenFradet:
I'm 👍 on dropping 2.10

suhasgaddam:
Ah, for some reason I thought it was only to bump sbt-org-policies. You can still make forward progress on converting to freestyle with keeping the sbt version to 0.13.16 and bumping sbt-org-policies.

It gets interesting because sbt-org-policies itself has a dependency on github4s(currently the cats 0.9 version due to another issue), but that dependency will not collide with the "real" project.

In terms of dropping 2.10(specifically the consequences for sbt-org-policies), I'll defer to @juanpedromoreno. We'll eventually stop publishing sbt-org-policies for 0.13.16, but I have no idea for that timeline.

juanpedromoreno:
Let's close this since other PRs have upgraded the build.

BenFradet:
unless I'm mistaken it's still on 0.13.x, no?

juanpedromoreno:
Yeah, that's true. IMHO, let's do it after this release, potentially removing Scala 2.10 support.

PR #190, opened 2018-01-09
Fixes Travis file

juanpedromoreno:
It seems the precise environment has been decommissioned...

https://github.com/47deg/github4s/commit/d49b540c247b7aea5ad2f931fa9495fbb7c56408 will fix the issue? No.

The latest commits, besides the files header update, bring build pipelines in Travis, which makes lightweight builds based on the Scala Version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: scala
 
-dist: precise
-
 scala:
 - 2.10.6
 - 2.11.11
@@ -20,7 +18,8 @@ install:
 - rvm use 2.2.8 --install --fuzzy
 - gem update --system
 - gem install sass
-- gem install jekyll -v 3.2.1
+- gem install ruby_dep -v 1.3.1
+- gem install jekyll -v 3.4.3
 - gem install jekyll-redirect-from
 
 before_script:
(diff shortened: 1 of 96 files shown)


BenFradet:
it works! 🎆

juanpedromoreno:
So I guess it's ready for review :)

.travis.yml
```diff
@@ -20,19 +34,49 @@ install:
...
+
+jobs:
+  include:
+    - stage: deploy
+      script:
+        - if [ "$TRAVIS_BRANCH" = "master" -a "$TRAVIS_PULL_REQUEST" = "false" ]; then
+            if grep -q "SNAPSHOT" version.sbt; then
+              sbt ++$TRAVIS_SCALA_VERSION publish;
+            else
+              if [ "$SCALAENV" = "all" ]; then
```
BenFradet: isn't this always the case due to the `env` on line 79?

fedefernandez: Yes, it should be

juanpedromoreno:
I'm not 100% sure, but it seems that the decommissioned precise environment was not working due to unknown reasons, but it was working fine locally.

Therefore I've been following a trial and error strategy until it works :(

PR #204, opened 2018-04-10
Change param so we are filtering by label

### LINKED ISSUE #203: Fix LabelParam
Looks like `LabalParam` is adding a `'state'` param rather than a `'label'`.

https://github.com/47deg/github4s/blob/b166c8706999191e67128f09744793070811c905/github4s/shared/src/main/scala/github4s/free/domain/SearchParam.scala#L70-L71

I am pretty new to this whole Scala thing, so I am not sure how this should be tested, but I'll update the PR with any suggestions.

drwlrsn:
Resolves #203.

### DIFF
--- a/github4s/shared/src/test/scala/github4s/utils/TestData.scala
+++ b/github4s/shared/src/test/scala/github4s/utils/TestData.scala
@@ -74,6 +74,7 @@ trait TestData extends DummyGithubUrls {
   val nonExistentSearchQuery = "nonExistentSearchQueryString"
   val validSearchParams = List(
     OwnerParamInRepository(s"$validRepoOwner/$validRepoName"),
+    LabelParam(label = "enhancement"),
     IssueTypeIssue,
     SearchIn(Set(SearchInTitle))
   )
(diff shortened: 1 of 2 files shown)


drwlrsn:
@BenFradet Just updated _TestData.scala_

> Thanks for catching the bug!

Hey my pleasure. Thanks for making a dope library. I am just starting to learn a bit of Scala and using this in a little toy project has been super clear and easy.

github4s/shared/src/test/scala/github4s/utils/TestData.scala
```diff
@@ -74,6 +74,7 @@ trait TestData extends DummyGithubUrls {
   val nonExistentSearchQuery = "nonExistentSearchQueryString"
   val validSearchParams = List(
     OwnerParamInRepository(s"$validRepoOwner/$validRepoName"),
+    LabelParam(label = "bug"),
```
BenFradet: Unfortunately this will fail since this test is used to retrieve the issue #40 which is labelled enhancement, that's what I meant in my review sorry if I wasn't clear.

drwlrsn: Oh hah!

BenFradet:
[0.18.4](https://github.com/47deg/github4s/releases/tag/v0.18.4) has been published and contains your fix :+1:

PR #207, opened 2018-06-11
Add support for the add labels endpoint

BenFradet:

PR #208, opened 2018-06-13
Bump Scala version to 2.12.6 in travis.yml

BenFradet:
Trying to solve the test flakiness from #207.

I'll dig deeper in the tests themselves if this doesn't solve it.


BenFradet:
doesn't seem to have solved it, I'll look into the cats effect js tests

PR #210, opened 2018-06-16
Add support for the remove label endpoint

BenFradet:


github4s/jvm/src/test/scala/github4s/utils/MockGithubApiServer.scala
```diff
@@ -885,6 +885,32 @@ trait MockGithubApiServer extends MockServerService with FakeResponses with Test
...
+        s"/repos/$validRepoOwner/$validRepoName/issues/$validIssueNumber/labels/${validIssueLabel.head}")
+      .withHeader("Authorization", tokenHeader))
+    .respond(response.withStatusCode(okStatusCode).withBody(listLabelsValidResponse))
+
+  mockServer
+    .when(
+      request
+        .withMethod("DELETE")
+        .withPath(
+          s"/repos/$validRepoOwner/$validRepoName/issues/$invalidIssueNumber/lablels/${validIssueLabel.head}")
```
juanpedromoreno: There is a typo here, in the URL path. Lablels-> labels

github4s/shared/src/test/scala/github4s/integration/GHIssuesSpec.scala
```diff
@@ -93,6 +93,17 @@ trait GHIssuesSpec[T] extends BaseIntegrationSpec[T] {
     })
   }
 
+  "Issues >> RemoveLabel" should "return a list of labels" in {
```
juanpedromoreno: Should remove the label from the issue successfully?

BenFradet: I went with `return a list of removed labels` since it actually returns the label that was removed in a list, wdyt?

PR #214, opened 2018-07-08
Abstract away from cats-effect IO

### LINKED ISSUE #213: [cats-effect] `Capture` can be derived for anything that has a `Sync` instance
`Capture.capture` is equivalent to `Sync.delay`, so an instance of `Capture[F]` can be implemented for every `F[_] : Sync`.

BenFradet:
fixes #213


BenFradet:
Could you please have a look @juanpedromoreno @fedefernandez ?

juanpedromoreno:
👍

cats-effect/js/src/main/scala/github4s/cats/effect/AsyncHttpRequestBuilderExtensionJS.scala
```diff
@@ -0,0 +1,60 @@
...
+      def runEmpty(rb: HttpRequestBuilder[SimpleHttpResponse, F]): F[GHResponse[Unit]] =
+        runMapWrapper[Unit](rb, emptyResponse)
+
+      private[this] def runMapWrapper[A](
+          rb: HttpRequestBuilder[SimpleHttpResponse, F],
+          mapResponse: SimpleHttpResponse => GHResponse[A]): F[GHResponse[A]] =
+        fromFuture(runMap[A, F](rb, mapResponse))
+
+      /** Taken from Http4s **/
+      private def fromFuture[A](future: Eval[Future[A]])(implicit ec: ExecutionContext): F[A] =
```
kubukoz: Note that there was an issue about having this in cats-effect. Although it'd take a `=> Future[A]`, but that's a minor detail. Note that the suggested implementation would short-circuit in case of completed futures (wouldn't use `onComplete` on `Future.successful` or `Future.failed`): https://github.com/typelevel/cats-effect/issues/199#issuecomment-401273282

BenFradet: The problem is that the reference implementation uses [TrampolineEC](https://github.com/typelevel/cats-effect/blob/98015a490a2cf9f4952c437176f97d86f45d0848/core/js/src/main/scala/cats/effect/internals/TrampolineEC.scala) and I didn't feel like copying that over too.

PR #230, opened 2018-11-27
Introduce a way to distinguish different github errors

mikegirkin:
I discovered that there is not way to distinguish between different type of errors in the application using github4s.
This small PR introduces another type for the errors, so the consumer get more information to deal with:
- ability to distinguish between HTTP errors (ie. file not found) and other
- ability to deal with HTTP errors differently based on status code


mikegirkin:
@BenFradet 
Not exactly, I've added http code to be able to distinguish on the client between different situations with a valid response from github received.
`UnexpectedException` is used also in JS-specific part to recover from Future errors and save the message.

BenFradet:
Ah yes forgot about hose damned `Future`s, wdyt about renaming it to `UnsuccessfulHttpRequest`, Github being implied since it subclasses `GHException`?

BenFradet:
you need to rename the other 2 occurrences :+1:

PR #240, opened 2018-12-24
Add Edit gist api

aleksandr-vin:
Closes #238

Issue #238, opened 2018-12-22
Implement Edit a gist API


aleksandr-vin:
Edit a gist API is not implemented:
https://developer.github.com/v3/gists/#edit-a-gist

pepegar:
:wave: same as in #237, if you want to make the contribution, I'm happy to help you :)

juanpedromoreno:
The project has a great contribution guide http://47deg.github.io/github4s/contributing.html :)

PR #241, opened 2018-12-24
Update Ruby to latest 2.3.8 for Travis CI

aleksandr-vin:
Should fix broken CI builds (see #238 and #240)


aleksandr-vin:
Looks like now Travis is working, but there appears to be a lot of failed tests in _master_. What to do with them?

BenFradet:
The tests are failing because the github token is only available to maintainers, I'll try to find the time to run them manually :+1:

PR #251, opened 2019-02-18
GetFollowing added to the algebra

mfirry:
See [documentation](https://developer.github.com/v3/users/followers/#list-users-followed-by-another-user).


mfirry:
Done... I think :)

Issue #255, opened 2019-03-01
Wrong API endpoint for listUserRepos


juli1:
In `listUserRepos` the API called is `users/$user/repos`. This seems to be the incorrect endpoint and be `/user/repos`

`listUserRepos` is documented to have a type parameter that can take the values "all", "public", "private",
   "forks", "sources" or "member" which seems to fit more or less with `/user/repos`. On the other hand, `users/$user/repos` does not even has the value `private` or `public` for the type. I guess there is a problem with the API being called.

BenFradet:
The implementation seems to correspond to https://developer.github.com/v3/repos/#list-repositories-for-a-user, feel free to reopen if I'm missing something.

PR #256, opened 2019-03-03
Add authenticated user list repo

juli1:
Add support for endpoint `/user/repos` as described in https://developer.github.com/v3/repos/#list-your-repositories .

- Fixes issue #255

### DIFF
--- a/github4s/shared/src/main/scala/github4s/api/Repos.scala
+++ b/github4s/shared/src/main/scala/github4s/api/Repos.scala
@@ -76,7 +76,7 @@ class Repos[C, M[_]](
     )
 
   /**
-   * List the repositories for a particular user
+   * List the public repositories for a particular user
    *
    * @param accessToken to identify the authenticated user
    * @param headers optional user headers to include in the request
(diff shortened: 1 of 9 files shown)


github4s/shared/src/main/scala/github4s/api/Repos.scala
```diff
@@ -100,6 +100,33 @@ class Repos[C, M[_]](
       pagination = pagination
     )
 
+  /**
+    * List repositories that the authenticated user has explicit permission to access
+    *
+    * @param accessToken to identify the authenticated user
+    * @param headers optional user headers to include in the request
+    * @param visiblility can be all, public, private
```
bilki:
```suggestion
    * @param visibility can be all, public, private
```

github4s/shared/src/main/scala/github4s/api/Repos.scala
```diff
@@ -100,6 +100,33 @@ class Repos[C, M[_]](
...
+     accessToken: Option[String] = None,
+     headers: Map[String, String] = Map(),
+     visibility: Option[String] = None,
+     affiliation: Option[String] = None,
+     `type`: Option[String] = None,
+     pagination: Option[Pagination] = Some(httpClient.defaultPagination)
+  ): M[GHResponse[List[Repository]]] =
+    httpClient.get[List[Repository]](
+      accessToken,
+      s"user/repos",
```
bilki:
```suggestion
      "user/repos",
```

github4s/shared/src/main/scala/github4s/api/Repos.scala
```diff
@@ -100,6 +100,33 @@ class Repos[C, M[_]](
...
+    * @param visiblility can be all, public, private
+    * @param affiliation comma-separated list of owner, collaborator or organization members
+    * @param `type` visibility of the retrieved repositories, can be "all", "public", "private",
+    * "forks", "sources" or "member"
+    * @return GHResponse[List[Repository]] the list of repositories for this user
+    */
+  def listAuthenticatedUserRepos(
+     accessToken: Option[String] = None,
+     headers: Map[String, String] = Map(),
+     visibility: Option[String] = None,
```
bilki: `visibility` and `affiliation` aren't being used within this request. Add them to the request query params like in `listCommits` endpoint. Keep in mind that you can't mix them with `type` param or github would return a 422 error.

PR #257, opened 2019-04-10
Make the integration tests skippable

BenFradet:
There are some unit tests in `ApiSpec` relying on the token being present so I ignored those as well

cc @SethTisue


SethTisue:
cool! I'd be happy to try this out in the community build once it's merged here. (or, I could try it out first, if you'd like.)

BenFradet:
I ran the tests locally without a token and it seemed to work but it'd be great to confirm it meets your expectations :+1:

PR #260, opened 2019-07-15
Adding merge_commit_sha field to PullRequest model

jdesiloniz:
This PR adds the `merge_commit_sha` to the `PullRequest` model. According to the GitHub API docs:

```
The value of the merge_commit_sha attribute changes depending on the state of the pull request. Before merging a pull request, the merge_commit_sha attribute holds the SHA of the test merge commit. After merging a pull request, the merge_commit_sha attribute changes depending on how you merged the pull request:

If merged as a merge commit, merge_commit_sha represents the SHA of the merge commit.
If merged via a squash, merge_commit_sha represents the SHA of the squashed commit on the base branch.
If rebased, merge_commit_sha represents the commit that the base branch was updated to.
```

This field can be really useful for creating and updating statuses by targetting the test commit sha instead of the latest one in a PR.

### DIFF
--- a/project/ProjectPlugin.scala
+++ b/project/ProjectPlugin.scala
@@ -124,10 +124,6 @@ object ProjectPlugin extends AutoPlugin {
       scalaVersion := V.scala212,
       crossScalaVersions := Seq(V.scala211, V.scala212),
       scalacOptions ~= (_ filterNot Set("-Xlint").contains),
-      scalacOptions ++= ((scalaBinaryVersion.value, getEnvVar("SCALAENV")) match {
-        case (x, env) if x.startsWith("2.12") && !env.contains("js") => Seq("-Ywarn-macros:after")
-        case _                         => Seq.empty
-      }),
       orgGithubTokenSetting := "GITHUB4S_ACCESS_TOKEN",
       resolvers += Resolver.bintrayRepo("hmil", "maven"),
       orgBadgeListSetting := List(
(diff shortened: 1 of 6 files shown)


fedefernandez:
@jdesiloniz I guess we need to disable that option for scala 2.11, something like:

```scala
scalacOptions ++= (scalaBinaryVersion.value match {
  case x if x.startsWith("2.12") => Seq("-Ywarn-macros:after")
  case _                         => Seq.empty
})
```

project/ProjectPlugin.scala
```diff
@@ -124,6 +124,10 @@ object ProjectPlugin extends AutoPlugin {
       scalaVersion := V.scala212,
       crossScalaVersions := Seq(V.scala211, V.scala212),
       scalacOptions ~= (_ filterNot Set("-Xlint").contains),
+      scalacOptions ++= ((scalaBinaryVersion.value, getEnvVar("SCALAENV")) match {
```
bilki: Is this still necessary?

version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.20.2-SNAPSHOT"
\ No newline at end of file
+version in ThisBuild := "0.20.2"
```
bilki: Are we going to release now?

jdesiloniz: @bilki I'm holding back from this, so we can decide the release timeline appropiatedly in a separate PR.

project/ProjectPlugin.scala
jdesiloniz: @bilki I've just launched a new build process in Travis without these, to verify if the stack size fix did the trick by itself... 🤔

jdesiloniz: @bilki yup, that did the trick! Thanks for the suggestion!!

PR #262, opened 2019-08-09
Scala 2.13 release, drop Scala 2.11 support

bilki:
Well, I've tried to cross-build github4s with Scala 2.13, and after tuning a bit SBT files and bumping dependencies, this is the current state:

- Macro paradise compiler plugin is no longer necessary for 2.13 (-Ymacro-annotations)
- We are missing the following dependencies (I'm going to open issues there asking about their expected roadmap to 2.13)
    - org.scalamock#scalamock-scalatest-support_2.13;3.6.0
    - fr.hmil#roshttp_sjs0.6_2.13;2.2.4
- Circe is dropping support for Scala 2.11 in 0.12 (and adding support for Scala 2.13), so we are forced to either drop Circe dependency or to drop Scala 2.11 support ourselves

I also bumped `sbt-org-policies` version to use `0.11.4`, with @juanpedromoreno changes to exclude `kind-projector` from default settings, which was producing version problems as well.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,5 +1,5 @@
 addSbtPlugin("com.eed3si9n"       % "sbt-buildinfo"            % "0.9.0")
-addSbtPlugin("com.47deg"          % "sbt-org-policies"         % "0.11.4-SNAPSHOT")
+addSbtPlugin("com.47deg"          % "sbt-org-policies"         % "0.12.0-M1")
 addSbtPlugin("com.47deg"          % "sbt-microsites"           % "0.9.2")
 addSbtPlugin("org.scala-js"       % "sbt-scalajs"              % "0.6.28")
 addSbtPlugin("org.portable-scala" % "sbt-scalajs-crossproject" % "0.6.1")
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,4 +1,5 @@
 addSbtPlugin("com.eed3si9n"       % "sbt-buildinfo"            % "0.9.0")
-addSbtPlugin("com.47deg"          % "sbt-org-policies"         % "0.9.4")
+addSbtPlugin("com.47deg"          % "sbt-org-policies"         % "0.11.3")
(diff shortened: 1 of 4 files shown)


bilki:
I bumped `sbt-org-policies` version, updated travis config file, and opened this PR for the last pending dependency that will allow us to build against 2.13:

https://github.com/hmil/RosHTTP/pull/88

Scalamock wasn't a stopper after all, bumped to 4.3.0 and it works fine with 2.13.

bilki:
By the way, compiling for 2.13 is issuing a ton of warnings about UTF-8 arrows, I will open another PR to migrate from those to plain two-character arrows.

project/plugins.sbt
```diff
@@ -1,4 +1,5 @@
-addSbtPlugin("com.eed3si9n"       % "sbt-buildinfo"            % "0.7.0")
-addSbtPlugin("com.47deg"          % "sbt-org-policies"         % "0.9.4")
-addSbtPlugin("org.scala-js"       % "sbt-scalajs"              % "0.6.26")
-addSbtPlugin("org.portable-scala" % "sbt-scalajs-crossproject" % "0.6.0")
+addSbtPlugin("com.eed3si9n"       % "sbt-buildinfo"            % "0.9.0")
+addSbtPlugin("com.47deg"          % "sbt-org-policies"         % "0.12.0-M1")
```
juanpedromoreno:
```suggestion
addSbtPlugin("com.47deg"          % "sbt-org-policies"         % "0.12.0-M2")
```

juanpedromoreno:
May I suggest removing the ScalaJS compatibility?

https://github.com/search?q=%22%25%25%25+github4s%22&type=Code

Unless my search is wrong, there are not many usages out there.

Thoughts?

BenFradet:
We can drop scalaz too :+1:

duanebester:
@bilki all arrows have now been migrated to plain two-character arrows

juanpedromoreno:
Superseded by https://github.com/47deg/github4s/pull/311

Issue #300, opened 2019-11-29
Remove list statuses test flakiness


BenFradet:
The following test:

https://github.com/47deg/github4s/blob/ea2285a1860d988dae9d26f54977ee47d08bb5fb/github4s/jvm/src/test/scala/github4s/unit/ApiSpec.scala#L403-L412

is extremely flaky

BenFradet:
seems to be solved, closing.

PR #301, opened 2019-11-29
Tagged test as integration

duanebester:
I found that setting the `GITHUB4S_ACCESS_TOKEN` allows the test to pass and not having it makes it fail, so I tagged the test as `Integration`

#300


fedefernandez:
@duanebester sorry I've been a bit out of this, could you please explain to me why this is working? Are those tests executed?

PR #330, opened 2020-01-24
Replace tut by mdoc and other clean ups

BenFradet:


github4s/src/main/scala/github4s/instances.scala
```diff
@@ -16,17 +16,19 @@
...
 import scala.concurrent.{ExecutionContext, Future}
 
-object implicits
+trait Instances
     extends FutureCaptureInstance
     with EvalInstances
     with IdInstances
     with FutureInstances
 
+object implicits1 extends Instances
```
fedefernandez: why you choose `implicits1` @BenFradet ?

BenFradet: I wanted to avoid the clash with `implicits` since these are internal I went for `implicits`, if you have an idea for a better name I'm :+1:

fedefernandez: A friend of us always said that the two hardest things in the engineering are cache invalidation and naming 😂

PR #344, opened 2020-02-13
Tagless final refactor

bond15:
Everything except mocked server unit tests.


github4s/src/test/scala/github4s/utils/FakeResponses.scala
```diff
@@ -574,6 +574,20 @@ trait FakeResponses {
       |[]
     """.stripMargin
 
+  val getSingleRefResponse =
```
BenFradet: since there is no `ApiSpec` anymore, this can be very much slimmed down, no?

github4s/src/main/scala/github4s/interpreters/UsersInterpreter.scala
```diff
@@ -0,0 +1,48 @@
...
+package github4s.interpreters
+
+import cats.Applicative
+import github4s.http.HttpClient
+import github4s.algebras.Users
+import github4s.GithubResponses.GHResponse
+import github4s.domain.{Pagination, User}
+import github4s.Decoders._
+
+class UsersInterpreter[F[_]: Applicative](
```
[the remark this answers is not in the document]
bond15: no need, forgot to delete

github4s/src/main/scala/github4s/algebras/Organizations.scala
```diff
@@ -14,73 +14,42 @@
...
-import io.circe.generic.auto._
+import github4s.domain.{Pagination, User}
 
-/** Factory to encapsulate calls related to Organizations operations  */
-class Organizations[M[_]](
-    implicit urls: GithubApiUrls,
-    httpClientImpl: HttpRequestBuilderExtension[M]) {
-
-  val httpClient = new HttpClient[M]
+abstract class Organizations[F[_]] {
```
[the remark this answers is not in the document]
bond15: No good reason haha, just a refactoring artifact. changed to trait.

github4s/src/main/scala/github4s/algebras/Auth.scala
```diff
@@ -77,21 +55,9 @@ class Auth[M[_]](
...
-            authorizeUrl.format(client_id, redirect_uri, scopes.mkString(","), state),
-            state),
-          statusCode = 200,
-          headers = Map.empty
-        )
-      )
-    )
-  }
+      scopes: List[String],
+      headers: Map[String, String] = Map()
```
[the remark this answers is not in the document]
bond15: Since the Auth interpreter never uses headers (or the http client) for `authorizeUrl`, I removed headers as a parameter for this method.

docs/docs/git_data.md
```diff
@@ -301,7 +305,7 @@ val createTag = Github(accessToken).gitData.createTag(
   "commit",
   Some(RefAuthor("2014-11-07T22:01:45Z", "rafaparadela", "developer@47deg.com")))
 
-createTag.exec[cats.Id]() match {
+createTag.toId match {
```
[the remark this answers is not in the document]
bond15: changed examples in documentation to unsafeRunSync

helpers for Id and Future are mentioned in docs.md

github4s/src/main/scala/github4s/interpreters/PullRequestsInterpreter.scala
```diff
@@ -0,0 +1,102 @@
...
+      repo: String,
+      filters: List[PRFilter],
+      pagination: Option[Pagination],
+      headers: Map[String, String] = Map()): F[GHResponse[List[PullRequest]]] =
+    client.get[List[PullRequest]](
+      accessToken,
+      s"repos/$owner/$repo/pulls",
+      headers,
+      filters.map(_.tupled).toMap,
+      pagination)
```
L-Lavigne: Named-argument style is inconsistent for this parameter in this file.

github4s/src/main/scala/github4s/interpreters/RepositoriesInterpreter.scala
```diff
@@ -0,0 +1,188 @@
...
+      Map(
+        "sha"    -> sha,
+        "path"   -> path,
+        "author" -> author,
+        "since"  -> since,
+        "until"  -> until
+      ).collect {
+        case (key, Some(value)) => key -> value
+      },
+      pagination
```
L-Lavigne: Named-argument style is inconsistent for this parameter in this file.

docs/docs/activity.md
```diff
@@ -112,4 +115,4 @@ As you can see, a few features of the activity endpoint are missing.
 
 As a result, if you'd like to see a feature supported, feel free to create an issue and/or a pull request!
 
-[activity-scala]: https://github.com/47deg/github4s/blob/master/github4s/shared/src/main/scala/github4s/free/domain/Activity.scala
+[activity-scala]: https://github.com/47deg/github4s/blob/master/github4s/src/main/scala/github4s/domain/Activity.scala
```
angoglez: It seems this url launches a 404 error

github4s/src/main/scala/github4s/Decoders.scala
```diff
@@ -255,4 +247,35 @@ object Decoders {
     }
   }
```
angoglez: Maybe it's better not to use reserved words to avoid using `` 

https://github.com/47deg/github4s/blob/146733dda87ad65582b7e5ef0ae013aebb77b552/github4s/src/main/scala/github4s/Decoders.scala#L154

github4s/src/main/scala/github4s/algebras/Issues.scala
```diff
@@ -14,19 +14,12 @@
...
-package github4s.api
+package github4s.algebras
 
 import github4s.GithubResponses.GHResponse
-import github4s.{GithubApiUrls, HttpClient, HttpRequestBuilderExtension}
-import github4s.free.domain._
-import github4s.util.URLEncoder
-import io.circe.syntax._
-import io.circe.generic.auto._
+import github4s.domain.{Comment, Issue, Label, Pagination, SearchIssuesResult, SearchParam, User}
```
angoglez:
```suggestion
import github4s.domain._
```

github4s/src/main/scala/github4s/http/Http4sSyntax.scala
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  implicit class HeadersOps(self: Headers) {
+    def toMap: Map[String, String] =
+      self.toList.map(header => (header.name.value, header.value)).toMap
+  }
+
+  implicit class RequestBuilderOps[R](self: RequestBuilder[R]) {
+
+    val toHeaderList: List[Header] = (self.headers.map({ kv =>
```
angoglez: unnecessary parenthesis inside map functions

github4s/src/main/scala/github4s/http/HttpClient.scala
```diff
@@ -0,0 +1,148 @@
...
+        .withHeaders(headers))
+
+  val defaultPagination   = Pagination(1, 1000)
+  val defaultPage: Int    = 1
+  val defaultPerPage: Int = 30
+  val resource: Resource[F, Client[F]] = BlazeClientBuilder[F](ec)
+    .withConnectTimeout(connTimeout)
+    .resource
+
+  private def buildURL(method: String) = urls.baseUrl + method
```
angoglez: It's a simple and private function but just to continue with the same signature as the others, maybe return type can be added to function definition

github4s/src/main/scala/github4s/interpreters/AuthInterpreter.scala
```diff
@@ -0,0 +1,79 @@
...
+      headers = Map("Authorization" -> s"Basic ${s"$username:$password".getBytes.toBase64}") ++ headers,
+      data = NewAuthRequest(scopes, note, client_id, client_secret)
+    )
+
+  override def authorizeUrl(
+      client_id: String,
+      redirect_uri: String,
+      scopes: List[String]): F[GHResponse[Authorize]] = {
+    val state = UUID.randomUUID().toString
+    val result: GHResponse[Authorize] =
```
angoglez: If this val it's not needed later, maybe `pure` can be done at the end of the contruction of the Either?

github4s/src/main/scala/github4s/interpreters/RepositoriesInterpreter.scala
```diff
@@ -0,0 +1,188 @@
...
+  override def listOrgRepos(
+      org: String,
+      `type`: Option[String],
+      pagination: Option[Pagination],
+      headers: Map[String, String] = Map()): F[GHResponse[List[Repository]]] =
+    client.get[List[Repository]](
+      accessToken,
+      s"orgs/$org/repos",
+      headers,
+      params = `type`.map(t => Map("type" -> t)).getOrElse(Map.empty),
```
angoglez:
```suggestion
      params = `type`.fold(Map.empty[String,String])(t => Map("type" -> t))
```
I think maybe each `map` followed by a `getOrElse` can be replaced by `fold`?

github4s/src/main/scala/github4s/GithubIOSyntax.scala
```diff
@@ -16,25 +16,27 @@
...
+  }
 
-}
+  implicit val idLiftIO: LiftIO[Id] = new LiftIO[Id] {
 
-object implicits extends InstancesAndInterpreters with FutureCaptureInstance
+    override def liftIO[A](ioa: IO[A]): Id[A] = ioa.unsafeRunSync()
+  }
+
+  implicit class IOOps[A](self: IO[A]) {
```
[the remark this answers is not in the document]
bond15: oo! Thanks for the pointer!

github4s/src/main/scala/github4s/http/Http4sSyntax.scala
```diff
@@ -0,0 +1,63 @@
...
+  implicit class RequestOps[F[_]: ConcurrentEffect](self: Request[F]) {
+    def withJsonBody[T](maybeData: Option[T])(implicit enc: Encoder[T]): Request[F] =
+      maybeData.fold(self)(
+        data =>
+          self
+            .withContentType(`Content-Type`(MediaType.application.json))
+            .withEntity(data.asJson.noSpaceNorNull))
+  }
+
+  implicit class JsonOps(self: Json) {
```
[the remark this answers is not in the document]
bond15: I made JsonOps and RequestBuilerOps value classes.

The compiler won't allow HeaderOps and RequestOps to be value classes

github4s/src/main/scala/github4s/GithubIOSyntax.scala
```diff
@@ -16,25 +16,27 @@
...
+
+    def toFuture(implicit liftIO: LiftIO[Future]): Future[A] = liftIO.liftIO(self)
+
+    def toId(implicit liftIO: LiftIO[Id]): Id[A] = liftIO.liftIO(self)
+  }
```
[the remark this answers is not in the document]
bond15: 1: Not sure what best practice is here for providing syntax. Open to suggestions

2: Can I have more context on that remark? I'm familiar with the idea that instances should obey laws, but how is that tested in scala (specifically for ConcurrentEffect)?

docs/docs/activity.md
bond15: yep! The location of the resources are changing. There is no longer a "shared" directory.

github4s/src/main/scala/github4s/Decoders.scala
bond15: This unfortunate naming is a result of trying to exactly match the names of the JSON fields in the responses.

bond15:
Hmm... Seems every few times the integration tests are run, "RemoveLabel" fails.. looking into this.

github4s/src/main/scala/github4s/Github.scala
```diff
@@ -16,55 +16,36 @@
...
+import github4s.algebras._
+import github4s.modules._
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.Duration
+
+class Github[F[_]: ConcurrentEffect](accessToken: Option[String], timeout: Option[Duration])(
+    implicit ec: ExecutionContext) {
+
+  val module: GithubAPIs[F] =
+    new GithubAPIv3[F](accessToken, timeout.getOrElse(Duration(1000l, MILLISECONDS)))
```
juanpedromoreno: Does it make sense if we make this val `private`?

github4s/src/main/scala/github4s/Github.scala
```diff
@@ -16,55 +16,36 @@
...
+
+  val users: Users[F]                 = module.users
+  val repos: Repositories[F]          = module.repos
+  val auth: Auth[F]                   = module.auth
+  val gists: Gists[F]                 = module.gists
+  val issues: Issues[F]               = module.issues
+  val activities: Activities[F]       = module.activities
+  val gitData: GitData[F]             = module.gitData
+  val pullRequests: PullRequests[F]   = module.pullRequests
+  val organizations: Organizations[F] = module.organizations
```
juanpedromoreno: Maybe it's not a big deal, but we also might want to make all these modules instantiations as `lazy`?

github4s/src/main/scala/github4s/http/RequestBuilder.scala
```diff
@@ -0,0 +1,49 @@
...
+
+  def putMethod: RequestBuilder[Res] = this.copy(httpVerb = Method.PUT)
+
+  def deleteMethod: RequestBuilder[Res] = this.copy(httpVerb = Method.DELETE)
+
+  def withHeaders(headers: Map[String, String]): RequestBuilder[Res] = this.copy(headers = headers)
+
+  def withParams(params: Map[String, String]): RequestBuilder[Res] = this.copy(params = params)
+
+  def withData(data: Res): RequestBuilder[Res] = this.copy(data = Option(data))
```
juanpedromoreno:
```suggestion
  def withData(data: Res): RequestBuilder[Res] = this.copy(data = Some(data))
```

github4s/src/main/scala/github4s/algebras/Issues.scala
```diff
@@ -14,19 +14,12 @@
...
-import github4s.util.URLEncoder
-import io.circe.syntax._
-import io.circe.generic.auto._
+import github4s.domain._
 
-/** Factory to encapsulate calls related to Issues operations  */
-class Issues[M[_]](implicit urls: GithubApiUrls, httpClientImpl: HttpRequestBuilderExtension[M]) {
-
-  val httpClient = new HttpClient[M]
+trait Issues[F[_]] {
```
juanpedromoreno: Out of curiosity, do we actually need these traits? Does it make sense if we directly define the algebra along with the implementation without needing to override these "interfaces"? Cannot think in any use case where these traits might be useful 🤔

github4s/src/main/scala/github4s/interpreters/IssuesInterpreter.scala
```diff
@@ -0,0 +1,167 @@
...
+import github4s.http.HttpClient
+import github4s.algebras.Issues
+import github4s.GithubResponses.GHResponse
+import github4s.domain._
+import java.net.URLEncoder.encode
+import github4s.Decoders._
+import github4s.Encoders._
+
+class IssuesInterpreter[F[_]](implicit client: HttpClient[F], accessToken: Option[String])
+    extends Issues[F] {
```
juanpedromoreno: Based on my comment above, we could remove this `extends`. If that makes sense, we could also provide a companion object. Something like this:

```scala
class Issues[F[_]](implicit client: HttpClient[F], accessToken: Option[String]) { ... }

object Issues {
  def apply[F[_]](implicit client: HttpClient[F], accessToken: Option[String]): Issues[F] = new Issues[F]
}
```

github4s/src/main/scala/github4s/modules/GithubAPIs.scala
```diff
@@ -0,0 +1,55 @@
...
+    extends GithubAPIs[F] {
+
+  implicit val client = new HttpClient[F](timeout)
+  implicit val at     = accessToken
+
+  override val users: Users[F]                 = new UsersInterpreter[F]
+  override val repos: Repositories[F]          = new RepositoriesInterpreter[F]
+  override val auth: Auth[F]                   = new AuthInterpreter[F]
+  override val gists: Gists[F]                 = new GistsInterpreter[F]
+  override val issues: Issues[F]               = new IssuesInterpreter[F]
```
juanpedromoreno: Following my suggestion:

```scala
val issues: Issues[F] = Issues[F]
```

github4s/src/test/scala/github4s/integration/GHActivitiesSpec.scala
```diff
@@ -16,19 +16,20 @@
...
 trait GHActivitiesSpec extends BaseIntegrationSpec {
 
   "Activity >> Set a thread subscription" should "return expected response when a valid thread id is provided" taggedAs Integration in {
     val response =
-      Github(accessToken).activities
-        .setThreadSub(validThreadId, true, false)
-        .execFuture(headerUserAgent)
+      Github[IO](accessToken).activities
+        .setThreadSub(validThreadId, true, false, headerUserAgent)
+        .toFuture
```
juanpedromoreno: Mhm, could we use `.attempt` instead?

Then the function: `testFutureIsLeft ` could perform the unsafeRunSync and check the result.

Note: this is valid just in case we want to avoid `Future` usage, as it's discouraged

github4s/src/main/scala/github4s/http/Http4sSyntax.scala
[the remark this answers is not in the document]
L-Lavigne: Yeah this is normal, as you probably know context bounds are syntactic sugar for an extra implicit parameter so they don't conform to the single-value requirement. This is fine, I like to use value classes when I can and leave others as-is.

github4s/src/main/scala/github4s/Github.scala
bond15: ah yes!

github4s/src/main/scala/github4s/algebras/Issues.scala
bond15: Functionally, that makes sense. The current implementation follows a tagless final pattern, thus separating the algebras and interpreters, but it only has one interpretation. 

I only have 2 weak arguments to keep the current definitions.
1. To follow tagless final pattern clearly
2. To cater to some other implementations in the future.

Idk what other interpreters would be interesting... 

- Maybe offering more than one choice of http client?

- support a Github V3 interface for the Github V4 API...
🤷‍♂ 

What do you think @rafaparadela ?

github4s/src/test/scala/github4s/integration/GHActivitiesSpec.scala
bond15: `GHResponse[A]` is a type alias for `Either[GHException, GHResult[A]]`

Using `.attempt` would nest the `Either`s
ex. 
response: IO[Either[Throwable, Either[GHException, GHResult[A]]]]

I could just run `.unsafeRunSync` on the `response` to check if it is `Left` or `Right` and then check the assertions.

Would it be desirable to run all these HTTP Requests synchronously instead of asynchronously?

github4s/src/main/scala/github4s/GithubIOSyntax.scala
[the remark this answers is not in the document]
rafaparadela: Basically, we want to encourage users of Github4s to move to `IO` or `Task`. But at the same time, given that in the previous versions all the examples and tests we defined in terms of `Future` and `Id`, we considered convenient to provide instances of `LiftIO[Future]` and `LiftIO[Id]`.
I'm in favor of rewriting the examples of the Microsite to have them using `iO`, and only in a separate section, we can mention, that for those who want to keep using `Future` (not recommended) we are providing `LiftIO[Future]`.

github4s/src/main/scala/github4s/algebras/PullRequests.scala
```diff
@@ -67,52 +59,38 @@ class PullRequests[M[_]](
...
       filters: List[PRFilter] = Nil,
-      pagination: Option[Pagination] = None): M[GHResponse[List[PullRequest]]] =
-    httpClient.get[List[PullRequest]](
-      accessToken,
-      s"repos/$owner/$repo/pulls",
-      headers,
-      filters.map(_.tupled).toMap,
-      pagination)
+      headers: Map[String, String] = Map(),
+      pagination: Option[Pagination] = None
```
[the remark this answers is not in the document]
rafaparadela: It used to be the last one if doesn't have pagination. But it's all right.

github4s/src/main/scala/github4s/interpreters/RepositoriesInterpreter.scala
rafaparadela: @angoglez Good catch!

github4s/src/main/scala/github4s/http/RequestBuilder.scala
rafaparadela: Why do you prefer `Some(data)` instead of `Option(data)`. I normally use the  `Option` constructor when the `A` comes externally.

```scala
  /** An Option factory which creates Some(x) if the argument is not null,
   *  and None if it is null.
   *
   *  @param  x the value
   *  @return   Some(value) if value != null, None if value == null
   */
  def apply[A](x: A): Option[A] = if (x == null) None else Some(x)
```

`Some(data))` is totally fine for me, but I'm just curious.

github4s/src/main/scala/github4s/algebras/Issues.scala
rafaparadela: One of the purposes of this PR is addressing the migration from _Free_ to _Tagless Final_, which normally expresses the algebras as type classes. And the modules as a composition of some of these algebras. 
The fact of having algebras as type classes allow generally to write different interpreters for runtime or testing. Another argument not applicable in Github4s, could be the capability to rewrite another interpretation of the algebra (using different HTTP client, or different DB driver, or whatever change) and having both of the co-existing, and respecting the trait as a contract).

I don't really see either a solid argument in favor of having these traits, beyond the fact of respecting the pattern and keep using this project a didactic example of it.

github4s/src/test/scala/github4s/integration/GHActivitiesSpec.scala
rafaparadela: As I mentioned in [this comment](https://github.com/47deg/github4s/pull/344#discussion_r381007470) I'm also inclined to rewrite the examples to use `IO` exclusively. 
And only a side note to mention that we don't recommend to use `Future` as the concurrent effect, but in the case the user wants to keep it, we are providing the `LiftIO[Future]`.

rafaparadela:
Thanks @anamariamv @angoglez @BenFradet @L-Lavigne @juanpedromoreno  for this amazing code review.

github4s/src/main/scala/github4s/algebras/Issues.scala
juanpedromoreno: Thanks both! We can move on as is, a lot of work here :)

I already approved, so LGTM!

github4s/src/main/scala/github4s/GithubIOSyntax.scala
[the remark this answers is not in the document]
BenFradet: 1. liftIO already provides its own syntax so people might already be used to that. What I'm trying to say is that defining a new syntax on top might not be needed.
2. cats defines laws for every type class (usually) that you can easily test when you define a new instance for a type class https://typelevel.org/cats/typeclasses/lawtesting.html, however there don't seem to be any defined for `LiftIO`: https://github.com/typelevel/cats-effect/tree/master/laws/shared/src/main/scala/cats/effect/laws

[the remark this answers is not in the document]
bond15: The Microsite examples are all using IO. In `docs.md` the `GithubIOSyntax` is mentioned and shown with examples.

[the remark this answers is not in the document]
bond15: As for if syntax should be provided:
I'm new to cats-effect and unaware of the existing syntax to lift IO to (Id / Future).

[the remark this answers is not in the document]
bond15: Thanks for the pointer about testing instances! However, I don't believe we are defining any new type class instances in this repo.

github4s/src/test/scala/github4s/integration/GHActivitiesSpec.scala
bond15: The examples are using IO. Question is, should the integration tests also use IO instead of Future.

juanpedromoreno: IMHO at some point it might be desired. However, it also could be tackled on a different PR ;)

github4s/src/main/scala/github4s/GithubIOSyntax.scala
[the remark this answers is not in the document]
BenFradet: These are new instances of `LiftIO`: https://github.com/47deg/github4s/pull/344/files/f85cce5f3ab4bef6762f2ad7b1d5d04d32a5d488#diff-90022f9f91efe2df61542a2d6996049dR26-R34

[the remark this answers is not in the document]
BenFradet: that's what I meant when defining new type class instances :+1:

Issue #340, opened 2020-02-05
Integrate a new API: Get the list of milestones in a project


rafaparadela:
As an active user of this library, we need to integrate another Github API to be able to get the list of the milestones for a given project, according to this specification: https://developer.github.com/v3/issues/milestones/#list-milestones-for-a-repository

PR #359, opened 2020-03-03
List milestone and List Labels in a project

anamariamv:
This PR solves #341 and #340

### DIFF
--- a/github4s/src/main/scala/github4s/interpreters/IssuesInterpreter.scala
+++ b/github4s/src/main/scala/github4s/interpreters/IssuesInterpreter.scala
@@ -144,10 +144,9 @@ class IssuesInterpreter[F[_]](implicit client: HttpClient[F], accessToken: Optio
   ): F[GHResponse[List[Label]]] =
     client.get[List[Label]](
       accessToken,
-      method = s"repos/$owner/$repo/labels",
-      headers,
-      Map(),
-      pagination
+      s"repos/$owner/$repo/labels",
+      headers = headers,
+      pagination = pagination
     )
 
   override def listLabels(
--- a/github4s/src/main/scala/github4s/interpreters/IssuesInterpreter.scala
+++ b/github4s/src/main/scala/github4s/interpreters/IssuesInterpreter.scala
@@ -136,6 +136,20 @@ class IssuesInterpreter[F[_]](implicit client: HttpClient[F], accessToken: Optio
   ): F[GHResponse[Unit]] =
     client.delete(accessToken, s"repos/$owner/$repo/issues/comments/$id", headers)
 
+  override def listLabelsRepository(
+      owner: String,
+      repo: String,
+      headers: Map[String, String],
+      pagination: Option[Pagination]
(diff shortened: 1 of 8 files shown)


github4s/src/main/scala/github4s/interpreters/IssuesInterpreter.scala
```diff
@@ -183,4 +197,23 @@ class IssuesInterpreter[F[_]](implicit client: HttpClient[F], accessToken: Optio
...
+      headers: Map[String, String]
+  ): F[GHResponse[List[Milestone]]] =
+    client.get[List[Milestone]](
+      accessToken,
+      s"repos/$owner/$repo/milestones",
+      headers,
+      pagination = pagination,
+      params = state.fold(Map.empty[String, String])(s => Map("state"  -> s)) ++
+        sort.fold(Map.empty[String, String])(s => Map("sort"           -> s)) ++
+        direction.fold(Map.empty[String, String])(d => Map("direction" -> d))
```
BenFradet:
```scala
List(state.map("state" -> _), sort.map("sort" -> _), direction.map("direction" -> _)).flatten.toMap
```

might be a bit better than concatenating maps? I'm not too sure but I feel like there might be a better way to do this 🤔

github4s/src/test/scala/github4s/utils/TestData.scala
```diff
@@ -49,19 +49,20 @@ trait TestData extends DummyGithubUrls {
...
-  val validPerPage       = 100
-  val validFilePath      = "README.md"
-  val invalidFilePath    = "NON_EXISTENT_FILE_IN_REPOSITORY"
-  val validDirPath       = "lib"
-  val validSymlinkPath   = "bin/some-symlink"
-  val validSubmodulePath = "test/qunit"
+  val validRepoOwner             = "47deg"
+  val invalidRepoOwner           = "invalid47deg"
+  val validRepoName              = "github4s"
+  val validRepoNameWithMilestone = "marlow"
```
BenFradet: not sure we want to use a private repo

rafaparadela: This info is private @anamariamv

github4s/src/test/scala/github4s/utils/TestData.scala
```diff
@@ -202,6 +203,44 @@ trait TestData extends DummyGithubUrls {
...
+    ),
+    open_issues = 6,
+    closed_issues = 1,
+    state = "open",
+    created_at = "2019-04-11T13:19:38Z",
+    updated_at = "2020-01-24T11:50:58Z",
+    due_on = "2020-03-31T07:00:00Z",
+    closed_at = None
+  )
+
```
rafaparadela: This `Marlow` info is private as well.

github4s/src/test/scala/github4s/utils/TestData.scala
anamariamv: Sorry, I used marlow because it returns a result with data and I forgot to change it, thanks.

PR #364, opened 2020-03-09
List cards

bond15:
Adds the functionality to list cards in a project by column id.

### DIFF
--- a/docs/docs/project.md
+++ b/docs/docs/project.md
@@ -90,4 +90,34 @@ The `result` on the right is the corresponding [List[Column]][column-scala].
 
 See [the API doc](https://developer.github.com/v3/projects/columns/#list-project-columns) for full reference.
 
-[column-scala]: https://github.com/47deg/github4s/blob/master/github4s/src/main/scala/github4s/domain/Column.scala
+[column-scala]: https://github.com/47deg/github4s/blob/master/github4s/src/main/scala/github4s/domain/Project.scala
+
+### Cards
+
+#### List project cards by column
+
+You can list the cards for a particular column with `listCards`; it takes as arguments:
+
+- `column_id`: column id for which we want to retrieve the cards.
+- `archived_state`: filters the project cards that are returned by the card's state. Can be one of `all`,`archived`, or `not_archived`. Default: `not_archived`
+- `pagination`: Limit and Offset for pagination, optional.
+- `header`: headers to include in the request, optional.
+
+To list the columns for project_id `8271018`:
+
+```scala mdoc:compile-only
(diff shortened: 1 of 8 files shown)


docs/docs/project.md
```diff
@@ -90,4 +90,34 @@ The `result` on the right is the corresponding [List[Column]][column-scala].
 
 See [the API doc](https://developer.github.com/v3/projects/columns/#list-project-columns) for full reference.
 
-[column-scala]: https://github.com/47deg/github4s/blob/master/github4s/src/main/scala/github4s/domain/Column.scala
+[column-scala]: https://github.com/47deg/github4s/blob/master/github4s/src/main/scala/github4s/domain/Project.scala
+
+### Cards
```
BenFradet: there is a missing entry for this section at the top of the page

BenFradet: l17-20

docs/docs/project.md
```diff
@@ -90,4 +90,34 @@ The `result` on the right is the corresponding [List[Column]][column-scala].
...
+
+#### List project cards by column
+
+You can list the cards for a particular column with `listCards`; it takes as arguments:
+
+- `column_id`: column id for which we want to retrieve the cards.
+- `archived_state`: filters the project cards that are returned by the card's state. Can be one of `all`,`archived`, or `not_archived`. Default: `not_archived`
```
BenFradet:
```suggestion
- `archived_state`: filters the project cards that are returned by the card's state. Can be one of `all`,`archived`, or `not_archived`. Default: `not_archived`, optional
```

bond15:
I think we may be triggering at 403 in the integrations tests by giving the wrong credentials too many times to the Auth api.

https://developer.github.com/v3/#authentication

Also, the integrations tests are sensitive to the `remove label` test. Not sure if this is a race condition between creating a label and then deleting it immediately.

docs/docs/project.md
```diff
@@ -124,4 +126,36 @@ The `result` on the right is the corresponding [List[Column]][column-scala].
...
+
+```scala mdoc:compile-only
+val listCards = Github[IO](accessToken).projects.listCards(
+    column_id = 8271018,
+    headers = Map("Accept" -> "application/vnd.github.inertia-preview+json"))
+val response = listCards.unsafeRunSync()
+response match {
+  case Left(e) => println(s"Something went wrong: ${e.getMessage}")
+  case Right(r) => println(r.result)
+}
```
BenFradet:
```suggestion
response.result match {
  case Left(e) => println(s"Something went wrong: ${e.getMessage}")
  case Right(r) => println(r)
}
```
I think this is all that's left :+1:

PR #370, opened 2020-03-13
Migrates from 47deg to 47degrees GH organization

juanpedromoreno:


docs/docs/activity.md
```diff
@@ -70,7 +70,7 @@ You can list the users starring a specific repository with `listStargazers`; it
 - `timeline`: whether or not to return the date at which point the user starred the repository.
 - `pagination`: Limit and Offset for pagination, optional.
 
-To list the stargazers of 47deg/github4s:
+To list the stargazers of 47degrees/github4s:
 
 ```scala mdoc:compile-only
 val listStargazers = Github[IO](accessToken).activities.listStargazers("47deg", "github4s", true)
```
BenFradet: do we need to modify the examples too?

BenFradet:
it might be that the token needs updating?

juanpedromoreno:
The token was updated, the problem is that the test data was pointing to the old organization. Now it's fixed and ready for review :)

PR #373, opened 2020-03-16
Require an http4s client to provide more flexibility

BenFradet:
This change avoids creating and tearing down a client on each call.

If this looks okay, I'll modify the docs as well.

cc @tovbinm

### DIFF
--- a/docs/docs/docs.md
+++ b/docs/docs/docs.md
@@ -13,21 +13,9 @@ appropriate scopes (i.e. if you want to create gists, your token will need to ha
 
 ## Github4s
 
-First things first, we'll need to import `github4s.Github` which is the entry point to the Github
-API in Github4s.
-
-```scala mdoc:silent
-import github4s.Github
-```
-```scala mdoc:invisible
-val accessToken = sys.env.get("GITHUB4S_ACCESS_TOKEN")
-```
-
-In order for Github4s to work, you'll need an appropriate implicit `ExecutionContext` in Scope.
-
 Github4s uses [Tagless Final encoding](https://typelevel.org/blog/2017/12/27/optimizing-final-tagless.html).
 
-Every Github4s API call returns an `F[GHResponse[A]]` where `F` has an instance of [cats.effect.ConcurrentEffect][cats-concurrent-effect].
+Every Github4s API call returns an `F[GHResponse[A]]` where `F` has an instance of [cats.effect.Sync][cats-sync].
 
 `GHResponse[A]` contains the result `A` given by Github (or an error) as well as the status code and
 headers of the response:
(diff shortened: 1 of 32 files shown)


cb372:
>This change avoids creating and tearing down a client on each call.

I haven't looked in detail but that might help with https://github.com/47degrees/sbt-microsites/issues/398

docs/docs/docs.md
```diff
@@ -40,103 +28,139 @@ final case class GHResponse[A](
...
+```scala mdoc:silent
+import github4s.Github
+val accessToken = sys.env.get("GITHUB4S_ACCESS_TOKEN")
+val user1 = Github[IO](httpClient, accessToken).users.get("rafaparadela")
 ```
 
 `user1` in this case is a `IO[GHResponse[User]]`.
 
-### Using `F[_]: cats.effect.ConcurrentEffect`
+### Using `F[_]: cats.effect.Sync`
```
cb372: The actual context bound in the sample code below is `: Async: ContextShift`

github4s/src/main/scala/github4s/Github.scala
```diff
@@ -16,19 +16,18 @@
...
+class Github[F[_]: Sync](
+    client: Client[F],
+    accessToken: Option[String]
 ) {
 
   private lazy val module: GithubAPIs[F] =
-    new GithubAPIv3[F](accessToken, timeout.getOrElse(Duration(1000L, MILLISECONDS)))
+    new GithubAPIv3[F](client, accessToken)
 
   lazy val users: Users[F]                 = module.users
```
cb372: Not related to this PR, but is there any reason for all this stuff to be lazy?

github4s/src/test/scala/github4s/integration/GHActivitiesSpec.scala
```diff
@@ -19,36 +19,41 @@ package github4s.integration
...
 
 trait GHActivitiesSpec extends BaseIntegrationSpec {
 
   "Activity >> Set a thread subscription" should "return expected response when a valid thread id is provided" taggedAs Integration in {
-    val response =
-      Github[IO](accessToken).activities
-        .setThreadSub(validThreadId, true, false, headerUserAgent)
-        .unsafeRunSync()
+    val response = client
+      .use { client =>
```
cb372: Maybe use a short name like `c` here, or rename the `client` in `BaseIntegrationSpec` to `clientResource` or something like, to avoid shadowing?

docs/docs/docs.md
BenFradet: This is because of the chosen client, if it were `BlazeClientBuilder` it would be `ConcurrentEffect`. Our constraint is actually `Sync`: https://github.com/47degrees/github4s/pull/373/files#diff-9a35e1db598f031cb40b6fcaf7266225R24. Do you think this needs additional information?

github4s/src/main/scala/github4s/Github.scala
BenFradet: I guess if you're using only one module, the others don't get evaluated? :thinking:

docs/docs/docs.md
BenFradet: or I could write the example as:

```scala
  def u1[F[_]: Sync](httpClient: Client[F]): F[GHResponse[User]] =
    Github[F](httpClient, accessToken).users.get("juanpedromoreno")
```

BenFradet: which might be closer to what people end up writing anyway

PR #384, opened 2020-03-22
Introduces custom GitHub configuration

### LINKED ISSUE #379: Support for private GitHub Enterprise installations
The base GitHub API URLs are hard-coded inside the `GithubAPIv3Config` class:
```scala
case class GithubAPIv3Config(
    baseUrl: String = "https://api.github.com/",
    authorizeUrl: String =
      "https://github.com/login/oauth/authorize?client_id=%s&redirect_uri=%s&scope=%s&state=%s",
    accessTokenUrl: String = "https://github.com/login/oauth/access_token"
)
```
which makes impossible to leverage **github4s** in those companies which use their private GitHub installations and obviously different base API URLs.

Currently it seems there's no way to override these URLs.

satorg:
Fixes #379

Adds an implicit `config` parameter to `Github` instance which allows to override the default configuration by injecting a custom implicit `GithubConfig` instance into a current scope:

```scala
val client: Client[IO] = ???
val accessToken: String = ???
implicit val myConfig = GithubConfig(baseUrl = "https://<whatever>/api/v3", ...)

val github = Github[IO](client, Some(accessToken)) // `myConfig` will be picked up while in the scope
```

### DIFF
--- a/docs/docs/docs.md
+++ b/docs/docs/docs.md
@@ -59,21 +59,6 @@ val user1 = Github[IO](httpClient, accessToken).users.get("rafaparadela")
 
 `user1` in this case is a `IO[GHResponse[User]]`.
 
-It is also possible to pass a custom GitHub configuration (e.g. a GitHub Enterprise server endpoint)
-by declaring an implicit `GithubConfig` value in the scope:
-
-```scala mdoc:silent
-import github4s.{Github, GithubConfig}
-
-implicit val config =
-  GithubConfig(baseUrl = "custom GitHub server API endpoint",  authorizeUrl = ???, accessTokenUrl = ???)
-
-val accessToken = sys.env.get("GITHUB4S_ACCESS_TOKEN")
-val github = Github[IO](httpClient, accessToken)
-
-```
-Please refer your GitHub server docs for exact configuration values.
-
 ### Using `F[_]: cats.effect.Sync`
 
 Any type with a `cats.effect.Sync` instance can be used with this example, such as
@@ -197,7 +182,30 @@ object ProgramEvalWithHeaders {
 }
 ```
 
+## Using github4s with GitHub Enterprise
+
+By default `Github` instances are configured for the [public GitHub][public-github] endpoints via a fallback
+`GithubConfig` instance which is picked up by the `Github` constructor if there's no other `GithubConfig` in the scope. 
+
+It is also possible to pass a custom GitHub configuration (e.g. for a particular [GitHub Enterprise][github-enterprise]
+server). To override the default configuration values declare a custom `GithubConfig` instance in an appropriate
+scope:
+```scala mdoc:silent
+import github4s.{Github, GithubConfig}
+
+implicit val config = GithubConfig(
+  baseUrl = ???,       // default: "https://api.github.com/"
+  authorizeUrl = ???,  // default: "https://github.com/login/oauth/authorize?client_id=%s&redirect_uri=%s&scope=%s&state=%s"
+  accessTokenUrl = ??? // default: "https://github.com/login/oauth/access_token"
+)
+
+val github = Github[IO](httpClient, "")
+```
+Please refer your GitHub Enterprise server docs for exact URL values for `baseUrl`, `authorizeUrl` and `accessTokenUrl`.
+
 [access-token]: https://github.com/settings/tokens
 [cats-sync]: https://typelevel.org/cats-effect/typeclasses/sync.html
 [monix-task]: https://monix.io/docs/3x/eval/task.html
 [http4s-client]: https://http4s.org/v0.21/client/
+[public-github]: https://github.com
+[github-enterprise]: https://github.com/enterprise
--- a/docs/docs/docs.md
+++ b/docs/docs/docs.md
@@ -59,6 +59,21 @@ val user1 = Github[IO](httpClient, accessToken).users.get("rafaparadela")
 
 `user1` in this case is a `IO[GHResponse[User]]`.
 
+It is also possible to pass a custom GitHub configuration (e.g. a GitHub Enterprise server endpoint)
+by declaring an implicit `GithubConfig` value in the scope:
+
+```scala mdoc:silent
+import github4s.{Github, GithubConfig}
+
+implicit val config =
+  GithubConfig(baseUrl = "custom GitHub server API endpoint",  authorizeUrl = ???, accessTokenUrl = ???)
+
+val accessToken = sys.env.get("GITHUB4S_ACCESS_TOKEN")
+val github = Github[IO](httpClient, accessToken)
+
+```
+Please refer your GitHub server docs for exact configuration values.
+
 ### Using `F[_]: cats.effect.Sync`
 
 Any type with a `cats.effect.Sync` instance can be used with this example, such as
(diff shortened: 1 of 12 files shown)


satorg:
Some questions and concerns I have got:

* I passed `config: GithubConfig` as an implicit parameters all the way from `Github` through `GithubAPIv3` and down to `HttpClient` where it is really used. But I am not completely sure it should be implicit everywhere. May be it is reasonable to make it implicit in the `Github` class only. 
* I found out that there's **application.conf** file in the **resources** directory with the same configuration defaults as the `GithubConfig.default` instance has. Is this configuration file used somewhere and shouldn't it be **reference.conf** instead of **application.conf** since **github4s** is supposed to be a library?
* The `Github` constructor accepts `accessToken` as a separate parameter, but shouldn't it be a part of the configuration object?

BenFradet:
>The Github constructor accepts accessToken as a separate parameter, but shouldn't it be a part of the configuration object?

No, the access token needs to be modified all the time, the default config will be fine for most people

>I found out that there's application.conf file in the resources directory with the same configuration defaults as the GithubConfig.default instance has. Is this configuration file used somewhere and shouldn't it be reference.conf instead of application.conf since github4s is supposed to be a library?

let's remove it, not sure why it's there

>I passed config: GithubConfig as an implicit parameters all the way from Github through GithubAPIv3 and down to HttpClient where it is really used. But I am not completely sure it should be implicit everywhere. May be it is reasonable to make it implicit in the Github class only.

I think top-level is enough, no need for implicits internally.

satorg:
@BenFradet I have updated the PR according to our conversations, please take a look.
I'm not sure if there's necessity to add some unit tests for these changes, but if you think the tests will be useful here please let me know.

github4s/src/main/scala/github4s/Github.scala
```diff
@@ -47,7 +47,6 @@ object Github {
   def apply[F[_]: Sync](
       client: Client[F],
       accessToken: Option[String] = None
-  ): Github[F] =
+  )(implicit urls: GithubConfig): Github[F] =
```
BenFradet:
```suggestion
  )(implicit config: GithubConfig): Github[F] =
```

github4s/src/main/scala/github4s/http/GithubConfig.scala
```diff
@@ -16,9 +16,14 @@
...
+
+object GithubConfig {
+  implicit val default: GithubConfig =
+    GithubConfig(
+      baseUrl = "https://api.github.com/",
+      authorizeUrl =
+        "https://github.com/login/oauth/authorize?client_id=%s&redirect_uri=%s&scope=%s&state=%s",
+      accessTokenUrl = "https://github.com/login/oauth/access_token"
+    )
+}
```
BenFradet: I think we'd need to move this up to the `github4s` package, wdyt?

github4s/src/main/scala/github4s/http/HttpClient.scala
```diff
@@ -27,9 +27,7 @@ import org.http4s.Request
 import org.http4s.circe.CirceEntityDecoder._
 import org.http4s.client.Client
 
-class HttpClient[F[_]: Sync](client: Client[F]) {
-
-  val urls: GithubAPIv3Config = GithubAPIv3Config()
+class HttpClient[F[_]: Sync](client: Client[F], val urls: GithubConfig) {
```
BenFradet:
```suggestion
class HttpClient[F[_]: Sync](client: Client[F], config: GithubConfig) {
```

satorg: Oh... `val urls` is used from outside of `HttpClient`. Do you think it is better to rename its usages as well?

BenFradet: yup :+1:

github4s/src/main/scala/github4s/Github.scala
satorg: fixed

github4s/src/main/scala/github4s/http/GithubConfig.scala
satorg: done

github4s/src/main/scala/github4s/http/HttpClient.scala
satorg: fixed

satorg:
@BenFradet,
I fixed PR according to your comments. Also I added some docs in the "Getting Started" section.
Although I'm not completely sure it is an appropriate place for that. Could you take a look please?

docs/docs/docs.md
```diff
@@ -59,6 +59,21 @@ val user1 = Github[IO](httpClient, accessToken).users.get("rafaparadela")
 
 `user1` in this case is a `IO[GHResponse[User]]`.
 
+It is also possible to pass a custom GitHub configuration (e.g. a GitHub Enterprise server endpoint)
```
BenFradet: I think this looks good but would need its own section title like `Using github4s with GitHub Enterprise` and need to be moved to the bottom.

docs/docs/docs.md
```diff
@@ -59,6 +59,21 @@ val user1 = Github[IO](httpClient, accessToken).users.get("rafaparadela")
...
+by declaring an implicit `GithubConfig` value in the scope:
+
+```scala mdoc:silent
+import github4s.{Github, GithubConfig}
+
+implicit val config =
+  GithubConfig(baseUrl = "custom GitHub server API endpoint",  authorizeUrl = ???, accessTokenUrl = ???)
+
+val accessToken = sys.env.get("GITHUB4S_ACCESS_TOKEN")
+val github = Github[IO](httpClient, accessToken)
```
BenFradet:
```suggestion
val customGithub = Github[IO](httpClient, "")
```

satorg:
@BenFradet, fixed, please take a look.
If you think that my English is a bit clumsy there then it is probably the case :)
So feel free to suggest a better wording please.

docs/docs/docs.md
```diff
@@ -182,7 +182,30 @@ object ProgramEvalWithHeaders {
...
+```scala mdoc:silent
+import github4s.{Github, GithubConfig}
+
+implicit val config = GithubConfig(
+  baseUrl = ???,       // default: "https://api.github.com/"
+  authorizeUrl = ???,  // default: "https://github.com/login/oauth/authorize?client_id=%s&redirect_uri=%s&scope=%s&state=%s"
+  accessTokenUrl = ??? // default: "https://github.com/login/oauth/access_token"
+)
+
+val github = Github[IO](httpClient, "")
```
BenFradet:
```suggestion
val github = Github[IO](httpClient, None)
```

docs/docs/docs.md
```diff
@@ -182,7 +182,30 @@ object ProgramEvalWithHeaders {
...
+server). To override the default configuration values declare a custom `GithubConfig` instance in an appropriate
+scope:
+```scala mdoc:silent
+import github4s.{Github, GithubConfig}
+
+implicit val config = GithubConfig(
+  baseUrl = ???,       // default: "https://api.github.com/"
+  authorizeUrl = ???,  // default: "https://github.com/login/oauth/authorize?client_id=%s&redirect_uri=%s&scope=%s&state=%s"
+  accessTokenUrl = ??? // default: "https://github.com/login/oauth/access_token"
+)
```
BenFradet:
```suggestion
  baseUrl = "",       // default: "https://api.github.com/"
  authorizeUrl = "",  // default: "https://github.com/login/oauth/authorize?client_id=%s&redirect_uri=%s&scope=%s&state=%s"
  accessTokenUrl = "" // default: "https://github.com/login/oauth/access_token"
)
```

PR #391, opened 2020-03-25
API: Create milestone

anamariamv:
Implement https://developer.github.com/v3/issues/milestones/#create-a-milestone

### DIFF
--- a/github4s/src/main/scala/github4s/algebras/Issues.scala
+++ b/github4s/src/main/scala/github4s/algebras/Issues.scala
@@ -314,4 +314,24 @@ trait Issues[F[_]] {
       headers: Map[String, String] = Map()
   ): F[GHResponse[List[Milestone]]]
 
+  /**
+   *
+   * @param owner repo owner
+   * @param repo repo name
+   * @param title The title of the milestone.
+   * @param state The state of the milestone. Either open or closed. Default: open
+   * @param description A description of the milestone.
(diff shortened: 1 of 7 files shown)


github4s/src/main/scala/github4s/algebras/Issues.scala
```diff
@@ -314,4 +314,24 @@ trait Issues[F[_]] {
       headers: Map[String, String] = Map()
   ): F[GHResponse[List[Milestone]]]
 
+  /**
+   *
```
BenFradet: missing doc

github4s/src/main/scala/github4s/algebras/Issues.scala
```diff
@@ -314,4 +314,24 @@ trait Issues[F[_]] {
...
+   * @param headers optional user headers to include in the request
+   * @return a GHResponse with the created Milestone
+   */
+  def createMilestone(
+      owner: String,
+      repo: String,
+      title: String,
+      state: Option[String],
+      description: Option[String],
+      due_on: Option[String],
```
BenFradet: I wonder if in these cases it wouldn't be better to supply a value of type date and format it ourselves, wdyt? :thinking:

juanpedromoreno: Yes, I agree, it looks like we'll need a [ZonedDateTime](https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html)

PR #440, opened 2020-04-08
Error refactor

BenFradet:
This PR aims to introduce more typed errors. it's still missing a few tests for the moment, hence the WIP.

### DIFF
--- a/docs/docs/docs.md
+++ b/docs/docs/docs.md
@@ -68,7 +68,7 @@ Any type with a `cats.effect.Sync` instance can be used with this example, such 
 object ProgramF {
   import cats.effect.Sync
   import github4s.Github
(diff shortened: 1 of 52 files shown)


BenFradet:
I'm currently hitting an issue with http4s detailed here: https://gist.github.com/BenFradet/bab012d44fe72b8bd6be73133c098863

Otherwise, it's ready to review :+1:

bond15:
Interesting, I think we also ran into that error mention in the gist.
Thanks for the updated error modeling! It will make life easier in Hoover :)

BenFradet:
fixed it by delegating the decoding to circe directly instead of http4s:

https://github.com/47degrees/github4s/pull/440/files#diff-b93ab619ec70a2dfcfee92460b1d5ac0R164-R167

PR #443, opened 2020-04-10
Uniformize pagination usage

BenFradet:


github4s/src/main/scala/github4s/algebras/Issues.scala
```diff
@@ -275,6 +283,7 @@ trait Issues[F[_]] {
       repo: String,
       number: Int,
       label: String,
+      pagination: Option[Pagination] = None,
```
dcsobral, line 286: Field was not added to scaladoc.

github4s/src/main/scala/github4s/interpreters/GitDataInterpreter.scala
```diff
@@ -30,6 +30,7 @@ class GitDataInterpreter[F[_]](implicit client: HttpClient[F], accessToken: Opti
       owner: String,
       repo: String,
       ref: String,
+      pagination: Option[Pagination] = None,
```
dcsobral, line 33: It needs to be used on the `get` call.

github4s/src/main/scala/github4s/interpreters/IssuesInterpreter.scala
```diff
@@ -32,6 +32,7 @@ class IssuesInterpreter[F[_]](implicit client: HttpClient[F], accessToken: Optio
   override def listIssues(
       owner: String,
       repo: String,
+      pagination: Option[Pagination] = None,
```
dcsobral, line 35: It needs to be used on the `get` call.

github4s/src/main/scala/github4s/interpreters/IssuesInterpreter.scala
```diff
@@ -96,6 +97,7 @@ class IssuesInterpreter[F[_]](implicit client: HttpClient[F], accessToken: Optio
       owner: String,
       repo: String,
       number: Int,
+      pagination: Option[Pagination] = None,
```
dcsobral, line 100: It needs to be used on the `get` call.

github4s/src/main/scala/github4s/interpreters/IssuesInterpreter.scala
```diff
@@ -155,6 +157,7 @@ class IssuesInterpreter[F[_]](implicit client: HttpClient[F], accessToken: Optio
...
```
dcsobral, line 160: The `pagination` parameter was added but not used anywhere. That is a problem in itself, but it also indicates another issue: it isn't being tested.

BenFradet:
Thanks for the review 👍 , went mindlessly through the algebras

BenFradet:
I think this is good to go now :+1:

dcsobral:
@BenFradet This will conflict with #462 , and introduces more cases of #449 . It will be easier if this is cleared of that issue, merged, and then #462 rebased on top of it.

BenFradet:
#462 is against the branch associated to this PR for that reason :+1:

PR #461, opened 2020-04-14
Add support for default headers

BenFradet:


github4s/src/main/scala/github4s/http/HttpClient.scala
```diff
@@ -144,19 +144,20 @@ class HttpClient[F[_]: Sync](client: Client[F], val config: GithubConfig) {
...
-  private def run[Req: Encoder, Res: Decoder](request: RequestBuilder[Req]): F[GHResponse[Res]] = {
+  private def run[Req: Encoder, Res: Decoder](request: RequestBuilder[Req]): F[GHResponse[Res]] =
     client
       .run(
         Request[F]()
           .withMethod(request.httpVerb)
           .withUri(request.toUri(config))
-          .withHeaders(request.toHeaderList: _*)
+          .withHeaders(
+            (config.headers.map { case (k, v) => Header(k, v) }.toList ++ request.toHeaderList): _*
```
dcsobral, line 154: I'd add a `toHeaderList` method to `GithubConfig` that does the left side of `++`.

PR #474, opened 2020-04-16
Docs readme

BenFradet:
@juanpedromoreno is this all that's required?


juanpedromoreno:
We probably will need to adapt the version in this file: https://github.com/47degrees/github4s/blob/master/docs/docs/index.md

juanpedromoreno:
... and update [this command alias](https://github.com/47degrees/github4s/blob/master/build.sbt#L2) to include the microsites docs to be updated:

```diff
-addCommandAlias("ci-docs", "project-docs/mdoc; headerCreateAll")
+addCommandAlias("ci-docs", "project-docs/mdoc; docs/mdoc; headerCreateAll")
```

See `mdoc` docs for reference: https://scalameta.org/mdoc/docs/installation.html

(the mdoc variables all provided by the [sbt-github-mdoc plugin](https://github.com/alejandrohdezma/sbt-github/blob/master/sbt-github-mdoc/src/main/scala/com/alejandrohdezma/sbt/github/mdoc/SbtGithubMdocPlugin.scala#L89-L105)).

juanpedromoreno:
mmm.. I have some doubts about my last comment:

```
sbt:docs> show mdocIn
[info] /github4s/docs/docs
sbt:docs> show mdocOut
[info] /github4s/docs/target/scala-2.12/resource_managed/main/jekyll
```

So presumably, the version will be updated when publishing the microsite. However, the modification in the command alias I suggested is not going to make any effect.

PR #492, opened 2020-05-19
List of releases

georgeorfanidi:


github4s/src/main/scala/github4s/algebras/Repositories.scala
```diff
@@ -271,6 +271,21 @@ trait Repositories[F[_]] {
       headers: Map[String, String] = Map()
   ): F[GHResponse[List[User]]]
 
+  /**
+   * List of releases
+   *
+   * @param owner of the repo
+   * @param repo name of the repo
+   * @param headers optional user headers to include in the request
```
BenFradet, line 279: :art: could you add scaladoc for the pagination param?

georgeorfanidi: Thank you very much!
 I have just added it.

PR #494, opened 2020-05-20
Get Blob support

georgeorfanidi:


github4s/src/test/scala/github4s/integration/ReposSpec.scala
```diff
@@ -121,6 +121,38 @@ trait ReposSpec extends BaseIntegrationSpec {
...
+    } yield gh.use(
+      _.gitData.getBlob(validRepoOwner, validRepoName, res.head.sha, headers = headerUserAgent)
+    )
+
+    blobResponse.isDefined shouldBe true
+
+    val responseBlob = blobResponse.get.unsafeRunSync()
+
+    testIsRight[BlobContent](responseBlob, r => r.content shouldBe fileContentO)
+    response.statusCode shouldBe okStatusCode
```
BenFradet, line 153: I think this can be simplified a bit: https://gist.github.com/BenFradet/4d5e85580ef986d7c39300d5690af867

georgeorfanidi: How about this approach?

BenFradet: yup that's cool too :+1:

PR #497, opened 2020-05-28
Get release by release Id

georgeorfanidi:


github4s/src/main/scala/github4s/algebras/Repositories.scala
```diff
@@ -271,6 +271,22 @@ trait Repositories[F[_]] {
...
+  /**
+   * Get a single release
+   *
+   * @param releaseId id of the release
+   * @param owner of the repo
+   * @param repo name of the repo
+   * @param headers optional user headers to include in the request
+   * @return a GHResponse with List[Release]
+   */
+  def getASingleRelease(
```
BenFradet, line 283: I think we can rename that to simply `getRelease`

georgeorfanidi: You are absolutely right, I just used the name which is given in github api documentation
https://developer.github.com/v3/repos/releases/#get-a-single-release

github4s/src/test/scala/github4s/integration/ReposSpec.scala
```diff
@@ -61,6 +62,39 @@ trait ReposSpec extends BaseIntegrationSpec {
...
+    val responseList: List[(Release, GHResponse[Option[Release]])] = responseResource
+      .use(identity)
+      .unsafeRunSync()
+
+    forAll(responseList) {
+      case (release, response) =>
+        testIsRight[Option[Release]](response, r => r should contain(release))
+        response.statusCode shouldBe okStatusCode
+    }
+  }
```
BenFradet, line 96: cool test :+1:

PR #523, opened 2020-06-30
Turn id fields into longs

chalenge:
Migrate all the id fields in github objects from `int` to use `long`. Solves an issue when the ids go over the max int such as `2170205650` which we saw as an id for the labels. I am not exactly sure of the `ids` are global to all repos but they do have really large numbers and hence decided to update all the other id fields for the client

Previous
```scala
final case class SomeGitObject(
  id: Int
  ....
)
```

Now
```scala
final case class SomeGitObject(
  id: Long
  ....
)
```


chalenge:
Will re-open this when I am fully done. Accidentally opened here instead of my fork 🤦

PR #530, opened 2020-07-21
Add check if user is a collaborator

zachkirlew:


github4s/src/main/scala/github4s/algebras/Repositories.scala
```diff
@@ -271,6 +271,26 @@ trait Repositories[F[_]] {
...
+   * organization members that are direct collaborators, organization members with access through team memberships,
+   * organization members with access through default organization permissions, and organization owners.
+   *
+   * @param owner of the repo
+   * @param repo name of the repo
+   * @param username Github username
+   * @param headers optional user headers to include in the request
+   * @return a unit GHResponse
+   */
+  def userIsCollaborator(
```
cb372, line 287: I think the return type is a bit confusing with respect to the method name. I would expect it to return `F[Boolean]`, but all the existing algebra methods I checked return a `GHResponse`, so that appears to be a design rule of the library.

@BenFradet WDYT?

zachkirlew: I did consider this but decided to maintain the `GHResponse` for two reasons:

1. Maintaining consistency with the library (as you mentioned)
2. Giving the user flexibility to handle other errors as they wish

github4s/src/main/scala/github4s/algebras/Repositories.scala
```diff
@@ -271,6 +271,26 @@ trait Repositories[F[_]] {
       headers: Map[String, String] = Map()
   ): F[GHResponse[List[User]]]
 
+  /**
+   * Get whether a user is a repository collaborator
+   *
+   * For organization-owned repositories, the list of collaborators includes outside collaborators,
+   * organization members that are direct collaborators, organization members with access through team memberships,
+   * organization members with access through default organization permissions, and organization owners.
+   *
```
BenFradet, line 280: Might be worth mentioning that a 404 means the user is not a collaborator, 204 means user is a collaborator.

BenFradet: or we might do the processing ourselves and return a `F[GHResponse[Boolean]]`

zachkirlew: I can change it to reflect this

cb372: How about this?

* keep it as `GHResponse[Unit]`
* document the possible status codes (204, 404) and what they mean
* rename the method to `getCollaborator`

zachkirlew: I would expect a method by that name to return `GHResponse[User]`

zachkirlew: WDYT of the latest changes returning `F[GHResponse[Boolean]]`

github4s/src/test/scala/github4s/integration/ReposSpec.scala
```diff
@@ -313,6 +313,57 @@ trait ReposSpec extends BaseIntegrationSpec {
     response.statusCode shouldBe notFoundStatusCode
   }
 
+  "Repos >> UserIsCollaborator" should "return true when response is successful" taggedAs Integration in {
```
BenFradet, line 316:
```suggestion
  "Repos >> UserIsCollaborator" should "return true when the user is a collaborator" taggedAs Integration in {
```

microsite/docs/repository.md
```diff
@@ -182,6 +183,28 @@ The `result` on the right is the corresponding [List[User]][user-scala].
 See [the API doc](https://developer.github.com/v3/repos/collaborators/#list-collaborators) for full
 reference.
 
+### Check if a user is a repository collaborator
+
+Returns whether a given user is a repository collaborator or not.
+
+204 No Content responses will produce a true result
+
+404 Not Found responses will produce a false result
```
BenFradet, line 192: We can remove that now :+1:

microsite/docs/repository.md
```diff
@@ -15,6 +15,7 @@ with Github4s, you can interact with:
   - [List user repositories](#list-user-repositories)
   - [List contributors](#list-contributors)
   - [List collaborators](#list-collaborators)
+  - [Check user is a repository collaborator](#check-if-a-user-is-a-repository-collaborator)
```
BenFradet, line 18:
```suggestion
  - [Check if a user is a repository collaborator](#check-if-a-user-is-a-repository-collaborator)
```

github4s/src/main/scala/github4s/algebras/Repositories.scala
```diff
@@ -271,6 +271,29 @@ trait Repositories[F[_]] {
...
+   * 204 No Content responses will produce a true result
+   * 404 Not Found responses will produce a false result
```
BenFradet, line 282:
```suggestion
```

github4s/src/main/scala/github4s/algebras/Repositories.scala
```diff
@@ -271,6 +271,27 @@ trait Repositories[F[_]] {
...
   ): F[GHResponse[List[User]]]
 
+  /**
+   * Get whether a user is a repository collaborator
+   *
+   * For organization-owned repositories, the list of collaborators includes outside collaborators,
+   * organization members that are direct collaborators, organization members with access through team memberships,
+   * organization members with access through default organization permissions, and organization owners.
+   *
+   *
```
BenFradet, line 281:
```suggestion
```

github4s/src/test/scala/github4s/integration/ReposSpec.scala
```diff
@@ -313,6 +313,57 @@ trait ReposSpec extends BaseIntegrationSpec {
...
+    testIsRight[Boolean](response, r => r should be(true))
+    response.statusCode shouldBe noContentStatusCode
+  }
+
+  it should "return false when the user is not a collaborator" taggedAs Integration in {
+    val response = clientResource
+      .use { client =>
+        Github[IO](client, accessToken).repos
+          .userIsCollaborator(
+            validRepoName,
```
cb372, line 338: I think this should be `validRepoOwner`?

BenFradet: I think that was meant to generate a 404.

github4s/src/test/scala/github4s/integration/ReposSpec.scala
```diff
@@ -313,6 +313,57 @@ trait ReposSpec extends BaseIntegrationSpec {
...
+    testIsRight[Boolean](response, r => r should be(false))
+    response.statusCode shouldBe notFoundStatusCode
+  }
+
+  it should "return error when other errors occur" taggedAs Integration in {
+    val response = clientResource
+      .use { client =>
+        Github[IO](client, accessToken).repos
+          .userIsCollaborator(
+            validRepoName,
```
cb372, line 355: If we are trying to trigger an UnauthorizedError here, maybe use strings that makes that more obvious? `"different-owner"` and `"some-repo"` or something like that.

BenFradet: I don't think it will be easy to generate one as gh tends to just hide things behind 404s.

github4s/src/test/scala/github4s/integration/ReposSpec.scala
cb372: Doesn't the `invalidUsername` do that? We are trying to test the "correct owner, correct repo, non-collaborator user" case here.

zachkirlew: Yeah sorry that was meant to be `validRepoOwner`

github4s/src/test/scala/github4s/integration/ReposSpec.scala
zachkirlew: Hmm do you think it's worth removing this test then?

github4s/src/test/scala/github4s/integration/ReposSpec.scala
BenFradet: ah, true indeed :+1:

github4s/src/test/scala/github4s/integration/ReposSpec.scala
zachkirlew: could use an invalid access token?

zachkirlew:
`buildResponse` isn't called in `HttpClient.getWithoutResponse` which is the method used in `userIsCollaborator`?

Also integration tests are ignored in my local for some reason

zachkirlew:
> Trying to run the tests locally I found the following issue: [buildResponse](https://github.com/47degrees/github4s/blob/master/github4s/src/main/scala/github4s/http/HttpClient.scala#L168-L188) relies on http4s' `attemptAs` which expects a non-empty body so we will always get a left.

`buildResponse` isn't called in `HttpClient.getWithoutResponse` which is the method used in `userIsCollaborator`?

Also integration tests are ignored in my local for some reason

BenFradet:
> > Trying to run the tests locally I found the following issue: [buildResponse](https://github.com/47degrees/github4s/blob/master/github4s/src/main/scala/github4s/http/HttpClient.scala#L168-L188) relies on http4s' `attemptAs` which expects a non-empty body so we will always get a left.
> 
> `buildResponse` isn't called in `HttpClient.getWithoutResponse` which is the method used in `userIsCollaborator`?

`getWithoutResponse` calls `run` which calls `buildResponse`.

To run the tests locally you need a `GITHUB_TOKEN` env var: https://47degrees.github.io/github4s/contributing#token

zachkirlew:
@BenFradet Tests fixed

cb372:
The tests were running for 45 minutes. I've just cancelled them and restarted 🤞

Issue #549, opened 2020-09-16
Default value for optional field


loonydev:
Hi guys, during working on https://github.com/47degrees/github4s/pull/547 I saw that optional fields don't have default parameter. 

For example, for `issue.createLabel` user need to have instance of `domain.Label` with `(name:String, color:String)` but he must to create object like `Label(None,name, None, None, color, None)`

Current model
```
final case class Label(
    id: Option[Long],
    name: String,
    description: Option[String],
    url: Option[String],
    color: String,
    default: Option[Boolean]
)
```

proposal
```
final case class Label(
    id: Option[Long] = None,
    name: String,
    description: Option[String] = None,
    url: Option[String] = None,
    color: String,
    default: Option[Boolean] = None
)
```

So user can create label like `Label(name = "String",color = "")`. 

WDYT?

juanpedromoreno:
Sounds good to me, I know those optional parameters do not need to be at the end, but if they aren’t, then you will be required to use named arguments syntax for any following arguments, as you suggested:

```scala
Label(name = "String", color = "")
```

Another alternative to re-order the case class is providing a companion object:

```scala
final case class Label(
        id: Option[Long],
        name: String,
        description: Option[String],
        url: Option[String],
        color: String,
        default: Option[Boolean]
    )

  object Label {
    def apply(name: String, color: String): Label = this(None, name, None, None, color, None)
  }
```

loonydev:
We can move option to the end and it's reduce amount of code:
```scala
final case class Label(
    name: String,
    color: String,
    id: Option[Long] = None,
    description: Option[String] = None,
    url: Option[String] = None,
    default: Option[Boolean] = None
)
```
So you can use it 
```scala
Label("name","color")
```

loonydev:
Great, will do soon!

fedefernandez:
@loonydev my vote is for the @juanpedromoreno's proposal. I don't really like too much the default values in the arguments and it's usually considered as a bad practice.

loonydev:
Hi @fedefernandez,

Well I'm agreed with you about default values, but it's `Option` field. Developer knows what to expect from it. 
The right way it's create a few constructors for possible situation. 
For example, for labels it will be:
- Create label without description `Label(name,color)`
- Create label with description `Label(name, color, description)`
- Response with label `Label(*all_fields*)`

And if with label it's not too complex, so what about `Content` which could be a `File`, `Directory`, `Link` and so on.

[...]

Issue #526, opened 2020-07-08
Include "permissions" in domain objects


zachkirlew:
Domain objects such as `Repository` do not contain the permissions field. It would be great to have these available to avoid having to other work arounds to detect if a user has read/write/admin privileges

PR #553, opened 2020-09-22
Added default value

loonydev:
Hi guys,
Close https://github.com/47degrees/github4s/issues/549. 
Btw, some `Option` field already has `None` values. So It was mess. Some add `Option` only, some `Option = None`.

Also close https://github.com/47degrees/github4s/issues/526 by added
```scala
final case class Repository(
    ...
    permissions: Option[RepoPermissions] = None
)

final case class RepoPermissions(
    admin: Boolean,
    push: Boolean,
    pull: Boolean
)
```

### DIFF
--- a/github4s/src/main/scala/github4s/domain/Issue.scala
+++ b/github4s/src/main/scala/github4s/domain/Issue.scala
@@ -21,7 +21,6 @@ import java.time.ZonedDateTime
 final case class Issue(
     id: Long,
     title: String,
-    body: Option[String],
     url: String,
     repository_url: String,
     labels_url: String,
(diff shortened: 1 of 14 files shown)


github4s/src/main/scala/github4s/domain/Issue.scala
```diff
@@ -30,31 +29,32 @@ final case class Issue(
...
 
 final case class IssuePullRequest(
-    url: Option[String],
-    html_url: Option[String],
-    diff_url: Option[String],
-    patch_url: Option[String]
+    url: String,
+    html_url: String,
+    diff_url: String,
+    patch_url: String
```
BenFradet, line 57: it's highly likely that the options were there for a reason.

BenFradet (commented):
LGTM, except for my comment :+1:

github4s/src/main/scala/github4s/domain/Issue.scala
loonydev: `IssuePullRequest` used in one place `Issue` where it's already `Option` field. I have check the [Issue docs](https://docs.github.com/en/rest/reference/issues) and don't see cases with some of this fields.

Well, I can change it back, but looks like overhead.

BenFradet: Unfortunately, the documentation is not a reliable source of truth. we've been bit by issues like this in the past.

loonydev: Sure, return option

loonydev:
Hi @BenFradet, 
Do I need something else here?

BenFradet (approved):
Nope, thanks :+1:

PR #565, opened 2020-10-27
Repository model enriched: Archived status available now

nihirash:
We're using this library for making tech.map of our organisation - so we need filter out archived repositories. 

To make it possible I did small changes - I've enriched repository model to support "archived" status.  

If PR need some improvements - please write me - I'll do it.


BenFradet (approved):
Thanks :+1:

Issue #554, opened 2020-09-22
GithubApp Auth


loonydev:
Hi, 
As I understand, for now this library don't support [GithubApps authorization](https://docs.github.com/en/developers/apps/authenticating-with-github-apps#authenticating-as-a-github-app)? 

If so, I want to discuss the best way for implementation.

In my code I used `trait GithubAuth` with method `def getAuthHeader():String`. `GithubClient` taken instance of `GithubAuth` and call this function for token. In case of PAT it's return token from constructor, in case of private key it's looks like:
```scala
class GithubKeyAuth(privateKeyContent: String, appId: String) extends GithubAuth {

  implicit val clock: Clock = Clock.systemUTC
  // build PrivateKey instance from string
  private val privateKey = GithubKeyAuth.getPrivateKey(privateKeyContent)

 // generate valid JWT token using PrivateKey
  override def getAuthHeader(): String = {
    "Bearer " + Jwt.encode(JwtClaim({
      s"""{"iss":$appId}"""
    }).issuedNow.expiresIn(10), privateKey, JwtAlgorithm.RS256)
  }
}
```

WDYT?

BenFradet:
You're right we don't support github apps authorization.
However, I'm not sure about the best way to go about it as we don't want to handle jwt tokens directly in the library for example.

loonydev:
Agreed, maybe we can make `GithubClient` to accept instance of `GithubAuth`, so the user can use by default `GithubPatAuth`, in other cases they need to define own class. 

But if it's part of Github Auth flow user will do same code over and over.

loonydev:
So, do you have any suggestions how to implement it? I can do that, by I need to be sure that I don't waste a time.

BenFradet:
>maybe we can make GithubClient to accept instance of GithubAuth, so the user can use by default GithubPatAuth, in other cases they need to define own class.

I'm :+1: on that

loonydev:
But as I mention before, users will do same code over and over and not in the best way. It's like give them domain and say, use own http sender.

BenFradet:
We can add that code as documentation, it doesn't seem to me like the usage will be so huge that it would need to be part of the library.

loonydev:
Okay, next week I'm going to prepare this changes and will discuss it with real example.

kusaeva:
@loonydev Hi! Are you still planning to make a PR with this changes?

loonydev:
[...]

PR #567, opened 2020-11-05
make accessToken an io to use gh app's expiring tokens

reimai:
Hi. In my project I want to use Github App auth, which operates expiring tokens. So here is a proposed change to the insides of the api:
* move accessToken into the http client 
* make it a monad

The default Github constructor stays the same, it's accessToken is just wrapped into a pure version of AccessTokens.
I've read you don't want to [bring things like jwt on board](https://github.com/47degrees/github4s/issues/554), so the implementation of AccessTokens for apps is left for users.


github4s/src/main/scala/github4s/algebras/AccessTokens.scala
```diff
@@ -0,0 +1,13 @@
+package github4s.algebras
+
+import github4s.GHResponse
+
+/**
+ * Source of static or expiring github tokens
+ */
+trait AccessTokens[F[_]] {
```
BenFradet, line 8:
```suggestion
trait AccessToken[F[_]] {
```

github4s/src/main/scala/github4s/algebras/AccessTokens.scala
```diff
@@ -0,0 +1,13 @@
...
+
+  def withAccessToken[T](f: Option[String] => F[GHResponse[T]]): F[GHResponse[T]]
+}
+
+object AccessTokens {}
```
BenFradet, line 13:
```suggestion
```

github4s/src/main/scala/github4s/interpreters/StaticAccessTokens.scala
```diff
@@ -0,0 +1,17 @@
+package github4s.interpreters
+
+import github4s.GHResponse
+import github4s.algebras.AccessTokens
+
+/**
+ * A simple static version
+ */
+class StaticAccessTokens[F[_]](accessToken: Option[String]) extends AccessTokens[F] {
```
BenFradet, line 9:
```suggestion
class StaticAccessToken[F[_]](accessToken: Option[String]) extends AccessToken[F] {
```

BenFradet (commented):
I like the approach :+1: 

Do the integration tests work locally?

Also I feel like we need to adapt the existing docs and create some more for someone looking to extend access token like you.

reimai:
I've renamed AccessTkoens and added some docs, my english might not be perfect though.

I tried running integrations tests locally with my personal token, and some of the tests fail for an obvious reason (me not having enough permissions for the github4s repository). Others are fine.

github4s/src/main/scala/github4s/modules/GithubAPIs.scala
```diff
@@ -45,3 +44,9 @@ class GithubAPIv3[F[_]: Sync](
   override val projects: Projects[F]           = new ProjectsInterpreter[F]
 
 }
+
+object GithubAPIv3 {
+
+  def noAuth[F[_]: Sync](client: Client[F], config: GithubConfig) =
```
BenFradet, line 50: missing return type

microsite/docs/docs.md
```diff
@@ -10,6 +10,8 @@ permalink: docs
 
 In order to access the Github API, you will need to have [an access token][access-token] with the
 appropriate scopes (i.e. if you want to create gists, your token will need to have the gist scope).
+A personal token lives forever, github app's one has to be re-issued every hour. 
+Implementation of [github app][github_app] authentication involves 3rd party libraries like jwt and is left for the user.
```
BenFradet, line 14:
```suggestion

A personal token can be valid forever. However, the ones issued for [Github Apps][github-app] have to be re-issued every hour. 
To solve that issue, github4s provides [AccessToken][access-token-scala] which you can hook into to provide authentication for your app using third-party libraries such as JWT implementations, for example.
You can find the default implementation for static tokens in [StaticAccessToken][static-access-token-scala].
```

microsite/docs/docs.md
```diff
@@ -240,6 +242,7 @@ headers which will be added to every request sent to the GitHub API. The user ag
 added by default.
 
 [access-token]: https://github.com/settings/tokens
+[github_app]: https://github.com/settings/apps
```
BenFradet, line 245:
```suggestion
[github-app]: https://github.com/settings/apps
[access-token-scala]: https://github.com/47degrees/github4s/blob/master/github4s/src/main/scala/github4s/algebras/AccessToken.scala
[access-token-scala]: https://github.com/47degrees/github4s/blob/master/github4s/src/main/scala/github4s/interpreters/StaticAccessToken.scala
```

BenFradet (approved):
Thanks a lot for your great contribution :+1:

juanpedromoreno (approved):
💯 Thanks for your time @reimai!

PR #595, opened 2021-01-12
Unbreak the build (fix #586 #597)

### LINKED ISSUE #586: Ignore the reviewers test
The reviewers test needs admin scope to succeed and the ci token doesn't seem to have it: https://github.com/47degrees/github4s/runs/1530047606?check_suite_focus=true#step:11:577

sloshy:
Fixes the build by solving two separate issues:
1. Ignores a failing test that was deemed safe to ignore (#586)
2. Removes the `Auth.newAuth` functionality as logging in with a username/password is now officially discontinued (#597)

This includes the requisite changes to documentation as well as a quick tweak to the `.gitignore` to properly filter out `metals.sbt` in all directories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,8 +34,7 @@ project/plugins/project/
 *.metals
 .bloop/
 .metals/
-project/.bloop/
-project/metals.sbt
+metals.sbt
 
 ### Vim ###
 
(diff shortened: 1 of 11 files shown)


sloshy:
Marked WIP as there seems to be another test failing. Not sure if it should hold up the fix but if there should be another issue for that, I'll happily remove the WIP prefix.

juanpedromoreno (approved):

sloshy:
Ok, after a lot of trial and error (Metals was not reporting any of the errors that github found) I've turned this PR into fixing the build entirely, and edited the description to match. Code coverage is slightly down but I don't think that's an issue since it's a result of one test being removed (as the feature was removed) and then one test being ignored (as the issue states).

BenFradet (approved):
:+1:

PR #599, opened 2021-01-20
Update mdoc, sbt-mdoc to 2.2.16

47erbot:
Updates 
* [org.scalameta:mdoc](https://github.com/scalameta/mdoc)
* [org.scalameta:sbt-mdoc](https://github.com/scalameta/mdoc)

 from 2.2.14 to 2.2.16.
[GitHub Release Notes](https://github.com/scalameta/mdoc/releases/tag/v2.2.16) - [Version Diff](https://github.com/scalameta/mdoc/compare/v2.2.14...v2.2.16)

I'll automatically update this PR to resolve conflicts as long as you don't change it yourself.

If you'd like to skip this version, you can just close this PR. If you have any feedback, just mention me in the comments below.

Configure Scala Steward for your repository with a [`.scala-steward.conf`](https://github.com/scala-steward-org/scala-steward/blob/8103d0289c2ae3f66a24a2e3cc4ca0f2257ca28a/docs/repo-specific-configuration.md) file.

Have a fantastic day writing Scala!

<details>
<summary>Ignore future updates</summary>

Add this to your `.scala-steward.conf` file to ignore future updates of this dependency:
```
updates.ignore = [ { groupId = "org.scalameta" } ]
```
</details>

labels: library-update, semver-patch

PR #601, opened 2021-01-26
Update Scala: 2.12.12/2.13.4, sbt-mdoc: 2.2.16

sloshy:
Supersedes #599. Updates `sbt-mdoc` and also updates Scala to the latest (as of today) versions for cross-building. The reason that branch was failing, I guess had something to do with the old version of Scala 2.12 we were cross-building for. As per: https://gitter.im/scalameta/mdoc?at=60107e009fa6765ef8db3dad

EDIT: The Scala 2.12 version was not able to be updated all  the way to 2.12.13, as that breaks compatibility with `sbt-coverage` as linked to below. For now, this is the lowest version that will allow `sbt-mdoc` to work without breaking other plugins.


sloshy:
Unfortunately this is still failing due to https://github.com/scoverage/sbt-scoverage/issues/319
sbt-coverage does not support the latest Scala versions. When this issue is updated and a solution is available, we can go forward with updating Scala versions.

sloshy:
For now I've downgraded Scala to 2.12.12 from 2.12.13. It's still an upgrade, but not all-the-way. 2.13.4 still works.

BenFradet (approved):
:+1:

PR #605, opened 2021-02-02
Add parent/source fields to Repository

### LINKED ISSUE #603: Repository doesn't contain a parent
I was trying to use github4s for a project I'm working on but wasn't able to because `Repository` doesn't seem to contain a parent even though it contains e.g. owner and organization (https://github.com/47degrees/github4s/blob/master/github4s/src/main/scala/github4s/domain/Repository.scala). Is there a specific reason for it? Could it be added?

sloshy:
Resolves #603 

Adds the `parent` and `source` fields to `Repository` which previously did not have them before. As the types are recursive and we define our own decoders, I decided to define these fields as a new type `RepositoryBase` which is exactly the same, only not recursive. The docs do not seem to indicate multiple levels of recursion so this is likely the more correct approach as well.

### DIFF
--- a/github4s/src/main/scala/github4s/Decoders.scala
+++ b/github4s/src/main/scala/github4s/Decoders.scala
@@ -108,7 +108,7 @@ object Decoders {
     }
   }
 
-  implicit val decodeRepository: Decoder[Repository] = {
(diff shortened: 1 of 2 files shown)


BenFradet (approved):
:+1:

PR #618, opened 2021-02-23
Fix Gitter link

kell18:
Hi! Super small change.


BenFradet (commented):
I think we were supposed to rename gitter rooms to match the repo coordinates at some point :thinking:

kell18:
Ah, got it, closing it then.

PR #662, opened 2021-06-09
Update http4s and cats-effect to CE3 era

### LINKED ISSUE #661: Update codebase to cats-effect 3
The CE3 update will also bring http4s to the 0.23 series

Daenyth:
Close #661 

This is blocked upstream, due to a behavior change in http4s.

- [ ] Handle the query parameter encoding issue
- [ ] Migrate/redesign the previous GithubIOSyntax helper
- [ ] Update docs


Daenyth:
Taken care of via a different PR

PR #665, opened 2021-06-14
Update README.md

PeterPerhac:
I just checked and dashing is no longer using github4s
see https://github.com/BenFradet/dashing/commit/83215f4b33dacec40f34385db753b7b67d46083a


BenFradet (approved):
ahah yes I migrated to graphql at some point :)

PR #680, opened 2021-08-11
634 double encoding breaks query params

josiah14:
## Problem Description

There was an issues with the building of some of the requests to the GitHub API which caused the query parameters to get encoded twice, first by the `java.net.URLEncoder` class and then by the `org.http4s.Uri.encode` function. Doing this would cause characters like `/` to get encoded as `%2F` and then as `%252F`, which the GitHub API would then decode only one layer down to `%2F`, causing inaccurate search results to come back both for the `Repos.searchRepos` and `Issues.searchIssues` functions.

## Applied Fix Description

Removed all usages of `java.net.URLEncoder` and allow all encoding to be done by `Http4sSyntax` for all requests to the Github API.

## Other Minor Changes

Some other minor improvements were made as part of this bugfix:

- Unix utilities work more reliably if files end in a newline character. This character was missing at the end of `logback.xml`, so it was added.
- In the testing utilities, `githubApiUrl` was not using TLS for its requests, so its value was altered to use `https` instead of `http`.
- In the `Repos.searchRepos` unit test spec, `"search/repositories"` was being supplied as the `url` to the mock `HttpClient`. This to me made the test seem less realistic and could be brittle if validation was ever added within the `HttpClient` to validate the string received is actually a `URL`  so as not to blindly accept just any string, so I changed the test to use the `githubApiUrl` value from the test utilities, instead.


github4s/src/test/scala/github4s/integration/ReposSpec.scala
```diff
@@ -471,6 +472,23 @@ trait ReposSpec extends BaseIntegrationSpec {
     response.statusCode shouldBe okStatusCode
   }
 
+  it should "successfully return results when a valid repo is provided using <owner>/<name> syntax" taggedAs Integration in {
+    val response = clientResource
+      .use { client =>
+        Github[IO](client, accessToken).repos
+          .searchRepos(Seq(validRepoOwner, validRepoName).mkString("/"), Nil)
```
BenFradet, line 479: :art: 
```suggestion
          .searchRepos(s"$validRepoOwner/$validRepoName", Nil)
```

BenFradet (approved):
Thanks :+1:

github4s/src/test/scala/github4s/integration/ReposSpec.scala
josiah14: https://github.com/47degrees/github4s/pull/680/commits/ae3310055d277a389336a3a6944eb4846064842f

@BenFradet I added removal of the import of `Seq` as well since it's no longer needed.

BenFradet (approved):

PR #683, opened 2021-08-20
Update branch and compare commits

reimai:
Adding apis to compare commits and update pull requests's branch from the base. 

The update-branch call is an [experimental api](https://docs.github.com/en/rest/overview/api-previews#update-a-pull-request-branch), so I've added an extra boolean parameter 'iAgreeToUseExperimentalApi' to match the api's usage of a specific header.


github4s/src/main/scala/github4s/interpreters/PullRequestsInterpreter.scala
```diff
@@ -160,4 +160,19 @@ class PullRequestsInterpreter[F[_]](implicit client: HttpClient[F]) extends Pull
...
+      owner: String,
+      repo: String,
+      pullRequest: Int,
+      iAgreeToUseExperimentalApi: Boolean,
+      expectedHeadSha: Option[String] = None,
+      headers: Map[String, String] = Map()
+  ): F[GHResponse[BranchUpdateResponse]] =
+    client.put[BranchUpdateRequest, BranchUpdateResponse](
+      s"repos/$owner/$repo/pulls/$pullRequest/update-branch",
+      headers ++ Option("Accept" -> "application/vnd.github.lydian-preview+json")
```
BenFradet, line 174: what happens if this header is not present?

BenFradet (commented):
I have a question regarding the header before I delve into the PR completely

github4s/src/main/scala/github4s/interpreters/PullRequestsInterpreter.scala
reimai: In this case an [unsupported media type is returned](https://docs.github.com/en/rest/reference/pulls#update-a-pull-request-branch): ```Status: 415 Unsupported Media Type```

BenFradet: Then I think we need to remove the `iAgree...` and document it's experimental in the scaladoc and example, wdyt?

reimai: Sure, that would be less confusing for users

reimai: Would you look at the changes?

github4s/src/main/scala/github4s/algebras/Repositories.scala
```diff
@@ -228,6 +229,22 @@ trait Repositories[F[_]] {
       headers: Map[String, String] = Map()
   ): F[GHResponse[List[Commit]]]
 
+  /**
```
BenFradet, line 232: maybe add what it does

BenFradet (commented):
Looks good :+1: , is there a way to add an integration test for update branch?

github4s/src/main/scala/github4s/algebras/Repositories.scala
reimai: sure

juanpedromoreno (approved):
LGTM, thanks for your time @reimai !

PR #690, opened 2021-09-15
Handle `default_branch` and `topics` in Repository

abestel:


abestel:
Looking in https://github.com/github/rest-api-description it is safe to assume that `default_branch` is required and always returned.
Topics is optional, and I thought it was nicer to expose a `List.empty[String]` instead of a `Option[List[String]]` in the domain object.

juanpedromoreno:
Thanks @abestel , would we need to update also the docs?

abestel:
> Thanks @abestel , would we need to update also the docs?

Thanks for checking my PR @juanpedromoreno :-)
I checked the documentation before submitting this change, and I don't think there's anything to update. The page for the repositories related actions is [here](http://47degrees.github.io/github4s/repository) and the model is referenced by linking directly to the actual source code, ie [here](https://github.com/47degrees/github4s/blob/main/github4s/src/main/scala/github4s/domain/Repository.scala).

I don't really understand why the Github Worflow failed by the way

github4s/src/main/scala/github4s/domain/Repository.scala
```diff
@@ -30,11 +30,13 @@ final case class RepositoryBase(
     updated_at: String,
     pushed_at: String,
     status: RepoStatus,
+    default_branch: String,
```
juanpedromoreno, line 33: _question_: Would it make sense to add the new fields at the end? Thinking about an easier way to upgrade for people using the existing model :)

juanpedromoreno (approved):
The pipeline failed for an unrelated issue, no worries about that, I checked your PR and everything looks right.

github4s/src/main/scala/github4s/domain/Repository.scala
abestel: I like to keep the required fields before the fields having a default value (hence the `default_branch` here and the `topics` all the way down) but I can definitely change it if you think it would be better :-)

juanpedromoreno: Your point makes sense to me too, let's keep it that way 👍

juanpedromoreno (approved):
Thanks so much for your time @abestel !

PR #693, opened 2021-09-16
Fix the build

abestel:
- Decoders was badly formatted after #690
- Integration tests were failing, it seems, because of the move from `master` to `main`


github4s/src/test/scala/github4s/utils/TestData.scala
```diff
@@ -132,14 +132,14 @@ trait TestData {
   val validTagSha   = "c3d0be41ecbe669545ee3e94d31ed9a4bc91ee3c"
 
   val validPullRequestFileSha      = "f80f79cafbe3f2ba71311b82e1171e73bd37a470"
-  val validPullRequestNumber       = 1
+  val validPullRequestNumber       = 637
```
abestel, line 135: There is a test trying to call the `update-branch` API on https://github.com/47degrees/github4s/pull/1 and it returns `422: Unprocessable Entity`... my best guess is that this is due to the PR pointing to `master`. 

https://github.com/47degrees/github4s/pull/637 looks like a good candidate to replace it

PR #747, opened 2022-01-11
Fix #746 - Expand FileComparison to Rename/Patch cases

### LINKED ISSUE #746: patch field being String in FileComparison causing compareCommit to crash
If the difference between two commits contains file difference that is only a rename (rename of file or moving file to another location), `patch` does not exist and therefore the parsing of the json fails. Instead `previous_filename` exists to indicate what name the file had before. This bug can easily be resolved by changing the type of patch to Option[String].

image

sloshy:
Fixes #746 

Took a slightly different approach than the one suggested. Because it seems like there are two distinct cases, I felt the user of the library should not have to know that these fields are mutually exclusive, and it should be instead clear from the models. If the `patch` field exists, it should be a `String` and not `Option[String]` since it sounds like that means there will also be no `previous_filename` field (and vice versa).

I've added two additional cases for `FileComparison` as `FileComparisonRename` and `FileComparisonPatch` that contain these two exclusive cases as separate models. It involves a little copy/paste but nothing egregious and I think this is better for end-users than the alternative anyway.

Either way, this is a breaking model change and should be noted when released.

EDIT: based on the below discussion, it seems the fields are not mutually exclusive but in the case of a rename there will always be `previous_filename` but not always a `patch`. I've modified the cases to match, and I've added methods to get these from the common supertype if they exist as `Option[String]`.

### DIFF
--- a/github4s/src/main/scala/github4s/domain/Repository.scala
+++ b/github4s/src/main/scala/github4s/domain/Repository.scala
@@ -371,9 +371,51 @@ object RepoUrlKeys {
       files: Seq[FileComparison] = Seq.empty
   )
 
-  sealed trait FileComparison
+  /**
+   * A file comparison that contains information on the changes to a file.
+   * There are two subtypes: `FileComparisonPatch` and `FileComparisonRename` that have different guarantees.
+   *
+   * * `FileComparisonPatch` guarantees that the `patch` field exists and has an optional `previous_filename` field.
+   * * `FileComparisonRename` guarantees that the `previous_filename` field exists and does not have a `patch` field.
+   *
+   * To get values from these fields, there are helper methods `getPatch` and `getPreviousFilename`, though
+   * it is recomended to match on your `FileComparison` value to determine which type it is, to remove ambiguity.
+   */
+  sealed trait FileComparison {
+    def sha: String
+    def filename: String
+    def status: String
+    def additions: Int
+    def deletions: Int
+    def changes: Int
+    def blob_url: String
+    def raw_url: String
+    def contents_url: String
+
+    /**
+     * Gets the contents of the `patch` field if it exists.
+     * To guarantee that the `patch` field is available, match this `FileComparison` value as a
+     * `FileComparison.FileComparisonPatch` type which always has this field.
+     */
+    def getPatch: Option[String]
+
+    /**
+     * Gets the contents of the `previous_filename` field if it exists.
+     * This field can sometimes appear in the case of a small file modification,
+     * but is guaranteed to appear in the event of a rename. To guarantee that
+     * this field is available, match this `FileComparison` value as a
+     * `FileComparison.FileComparisonRename` type which always has this field.
+     */
+    def getPreviousFilename: Option[String]
+  }
 
   object FileComparison {
+
+    /**
+     * Represents a file comparison where the only change was a rename.
+     * The `patch` field does not exist, and the `previous_filename` field is guaranteed to exist,
+     * containing the file's previous filename.
+     */
     final case class FileComparisonRename(
         sha: String,
         filename: String,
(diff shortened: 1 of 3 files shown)


sloshy:
In the event that these are not exclusive cases, I'm happy to make the change to them being `Option[String]` or maybe changing the cases around somewhat. Curious if there is a third case for "rename + patch" as well.

sloshy:
Seems they are exclusive according to: https://docs.github.com/en/rest/reference/commits#compare-two-commits

> For example, files with a `renamed` status have a previous_filename field showing the previous filename of the file, and files with a `modified` status have a patch field showing the changes made to the file.

In the examples below it looks like the status cannot be both `modified` and `renamed` and in the `modified` case it's not clear that `previous_filename` can still possibly appear.

github4s/src/main/scala/github4s/domain/Repository.scala
```diff
@@ -371,17 +371,34 @@ object RepoUrlKeys {
...
-      status: String,
-      additions: Int,
-      deletions: Int,
-      changes: Int,
-      blob_url: String,
-      raw_url: String,
-      contents_url: String,
-      patch: String
-  )
+  sealed trait FileComparison
```
fedefernandez, line 374: Would it make sense to add the fields in the trait too (as a `def`s)? I think it could be useful for handling the common fields

```scala
sealed trait FileComparison {
  def sha: String
  //...
}
```

rojsrojs:
> Seems they are exclusive according to: https://docs.github.com/en/rest/reference/commits#compare-two-commits
> 
> > For example, files with a `renamed` status have a previous_filename field showing the previous filename of the file, and files with a `modified` status have a patch field showing the changes made to the file.
> 
> In the examples below it looks like the status cannot be both `modified` and `renamed` and in the `modified` case it's not clear that `previous_filename` can still possibly appear.

The status cannot be both `modified` and `renamed`. However, the change can have both the fields `previous_filename` and `patch`. If you make a small modification and a rename at the same time, the status will be `rename` and have the `patch` field with the small change. If you make a big change, we will get two file comparisons and one of them will be `removed` and the other one `added` and the `previous_filename` won't appear in any of them. I suspect you get the status `modified` if you make a "medium" change, but I haven't tested that case.

github4s/src/main/scala/github4s/domain/Repository.scala
sloshy: Done. Also added a couple helper methods for `getPatch` and `getPreviousFilename` for the cases where you don't necessarily want to do pattern matching.

sloshy:
Thanks @rojsrojs for your inputs. I've changed the two cases as follows:

* Renamed to `FileComparisonModified` and `FileComparisonRenamed`
* For the `Modified` case, it has `patch` but no `previous_filename`
* For the `Renamed` case it has `previous_filename` and an optional `patch` field.
* Getters were added to the base trait so users don't have to pattern match to get these fields, but they return `Option[String]` without pattern matching so they are not guaranteed.

sloshy:
Just noticed and fixed an encoder/decoder ambiguity. Because the cases are separated now and can be parsed as each other, I've modified the encoder/decoder to take into account the expected `status` of the `Renamed` case. I've also changed the `Modified` case to be `NotRenamed` as I felt that made more sense given the possibility for it to be an added or deleted file.

github4s/src/main/scala/github4s/domain/Repository.scala
```diff
@@ -371,17 +371,83 @@ object RepoUrlKeys {
...
+    def blob_url: String
+    def raw_url: String
+    def contents_url: String
+
+    /**
+     * Gets the contents of the `patch` field if it exists, in the case that the file was modified.
+     * To guarantee that the `patch` field is available, match this `FileComparison` value as a
+     * `FileComparison.FileComparisonNotRenamed` type which always has this field.
+     */
+    def getPatch: Option[String]
```
fedefernandez, line 400: Would it make sense to remove the `getPatch` and `getPreviousFilename` fields from here? If the user wants these fields, they must match the concrete type. If they don't need those fields, they can use the supertype.

fedefernandez (approved):
Left a minor comment, but it looks good to me.

github4s/src/main/scala/github4s/domain/Repository.scala
sloshy: I think it's a lot less verbose to use these rather than do a proper pattern match for some cases. If I'm using this library and I want "the previous filename if it exists" then that's a single method call. Or if I want the patch "if it exists" that's also a single call. Takes up a lot less vertical space. The main use case of doing the pattern match would be if you're passing the file comparison to some downstream function and want to validate that it's the right subtype first, but I don't think it should be required if we can help it.

PR #789, opened 2022-04-27
Add new endpoint - List public organization events

necosta:
Implemented endpoint for https://docs.github.com/en/rest/activity/events#list-public-organization-events

Went with a "minimalist" approach mapping only core fields, in a future PR we can improve mapping the payload (per-event-type) fields.
https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types

Ready for review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,4 +28,4 @@ If you wish to add your library here please consider a PR to include it in the l
 
 Github4s is designed and developed by 47 Degrees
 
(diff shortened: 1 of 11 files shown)


fedefernandez (approved):
LGTM

josevi96 (approved):
LGTM!, I'm working on the [List repository events](https://docs.github.com/en/rest/reference/activity#list-repository-events) endpoint, maybe we should merge this PR first, thank you!

PR #811, opened 2022-08-10
add new endpoint - list repository events

josevi96:
New endpoint to list the events of a particular repository with `listPublicRepositoryEvents` using the owner and the repo's name.
Github docs:
https://docs.github.com/en/rest/activity/events#list-repository-events


josevi96:
Hi folks!, I was thinking about merging the PublicRepositoryEvent and the PublicOrganizationEvent as they are almost the same but as they are different entities in the github API I decided to separate them, what do you think?

github4s/shared/src/main/scala/github4s/Decoders.scala
```diff
@@ -265,6 +265,25 @@ object Decoders {
       )
     )
 
+  implicit val decodePublicRepositoryEvent: Decoder[PublicRepositoryEvent] =
```
fedefernandez, line 268: `PublicRepositoryEvent` model looks similar to `PublicOrganizationEvent`. Can you please check the API and see if we can reuse the model?

josevi96: Yes they share the same schema, we can merge both models into the same one

necosta (approved):
Looks good. Great contribution 💪

fedefernandez (approved):

PR #817, opened 2022-08-27
fix test assert returning success in case of decoding error

nicoulaj:
`testIsRight` actually passes tests with `DecodingError` as success, this fixes it


fedefernandez (approved):
Thanks!

juanpedromoreno (approved):
Thanks @nicoulaj

PR #830, opened 2022-09-23
Fix GithubPublicEvent schema

javimartinez:
After https://github.com/47degrees/github4s/pull/817 the tests for the following operations are failing:  

- List public organization events
- List repository events

That's because the expected schema in the code doesn't match with the JSON returned by the GitHub API. The field `repo.full_name` doesn't exist; instead, we can get the `repo.name` field. 

For further reference: https://docs.github.com/en/rest/activity/events#list-public-repository-events, https://docs.github.com/en/rest/activity/events#list-public-repository-events


franciscodr (approved):
Thanks, @javimartinez

